PR #10, opened 2015-10-26
Adding option to disable strict SSL verification

tehbilly:
I'm not a "javascript guy", but submitting a change to allow ignoring strict SSL checking (because corporate proxies woo).


maxogden:
nice, LGTM

maxogden:
published as 1.3.0

PR #13, opened 2016-03-03
Add custom 404 message

kevinsawicki:
Adds a custom 404 message that front-loads the version, platform, and arch information.

This is an attempt to better showcase there is no Electron version for certain platform/arch combinations such as 32-bit Mac.

Refs https://github.com/atom/electron/issues/4655

/cc @jlord :pear:

### DIFF
--- a/index.js
+++ b/index.js
@@ -53,15 +53,16 @@ module.exports = function download (opts, cb) {
         if (err) return cb(err)
         debug('downloading zip', url, 'to', tmpdir)
         var nuggetOpts = {target: filename, dir: tmpdir, resume: true, verbose: true, strictSSL: strictSSL, proxy: proxy}
-        nugget(url, nuggetOpts, function (err) {
-          if (/404/.test(err)) {
+        nugget(url, nuggetOpts, function (errors) {
+          if (errors) {
+            var error = errors[0]
             if (symbols) {
-              err.message = 'Failed to find Electron symbols v' + version + ' for ' + platform + '-' + arch + ' at ' + url
+              error.message = 'Failed to find Electron symbols v' + version + ' for ' + platform + '-' + arch + ' at ' + url
             } else {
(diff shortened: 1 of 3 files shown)


jlord:
:wave:

maxogden:
@kevinsawicki HELLO. Only thing that I don't like about this is its a breaking change since the Error from nugget is no longer getting passed to the callback, meaning that the `.stack` property and `.message` and whatever else will be different on your new Error object. I wonder if theres a way to do it that doesn't modify the nugget error?

kevinsawicki:
@maxogden thanks for the feedback, I pushed a new commit that re-assigns the message instead of creating a new error, does that sound okay to you?

maxogden:
@kevinsawicki so I just remembered that because this just forwards the error from `nugget`, and `nugget` returns an array of errors (because nugget also accepts an array of URLs, it seemed like a good idea at the time), we should actually just do:

```
var error = errors[0]
error.message = 'your message'
cb(error)
```

Does that make sense? In this case we know there is only one error in the errors array because we only pass one url in, so errors[0] should always work

maxogden:
alternative is we could change nugget to either return only 1 error, since people probably arent even handling multiple errors in their code, or have it return a single error if only a single url was given. then it would prevent us having to do `errors[0]` here

kevinsawicki:
@maxogden I went with your first suggestion, plucking the first error out, seems like a good approach vs. breaking changes in nugget.

maxogden:
ok cool i can live with this, gonna bump major and then upgrade the dep in electron-prebuilt, it will get released the next time electron is released

kevinsawicki:
Awesome, thanks

maxogden:
ok published, also i added https://github.com/electron-userland/electron-download/commit/245e5ef368498535e3df1e3d23a6c5b8cf310e5f

kevinsawicki:
> ok published, also i added 245e5ef

Thanks, yeah, I accidentally deleted the regex check.

PR #14, opened 2016-03-12
Allow customization of URL directory and filename

quicksnap:
Internally we would like to use our own mirror which does not conform to the `ELECTRON_MIRROR`/`VERSION`/`NAME`. Would you accept a PR like this?


maxogden:
Sure, can you add some docs though?

quicksnap:
Yeah totally! Thanks. I'll add them soon.

quicksnap:
I've updated the PR with documentation. Look good?

PR #25, opened 2016-09-07
Add SHASUMS256.txt checking

### LINKED ISSUE #19: Verify checksums of Electron zip files
Checksums should be provided by the Electron project. The verification process looks like this:

1. Download and parse the checksum file(s). If a 404 is encountered, skip steps 3 & 4
2. Download the zip
3. Generate a checksum against the downloaded zip
4. If the checksums do not match, throw an exception

Blocked on https://github.com/electron/electron/pull/6588

malept:
When an Electron release has a `SHASUMS256.txt` available, it will parse that file and verify that the checksum for the given file matches what is generated by the `sumchecker` module using the contents of the file. If the file does not match, it is deleted from the Electron cache and an error is thrown.

This PR also replaces `mkdirp` and `mv` with `fs-extra`, since I had to use `fs.unlink` anyway.

CC: @kevinsawicki @zeke 

Fixes #19.

### DIFF
--- a/test_404.js
+++ b/test_404.js
@@ -2,6 +2,7 @@
 
 const download = require('./')
 
+console.log('404 test')
 download({
   version: '0.25.1',
   arch: 'ia32',
@@ -12,5 +13,4 @@ download({
     throw Error('Download did not throw an error with a custom 404 message: ' + err.message)
   }
   console.log('OK! got expected error:', err.message)
-  process.exit(0)
 })
--- a/index.js
+++ b/index.js
@@ -1,14 +1,15 @@
 'use strict'
 
 const debug = require('debug')('electron-download')
+const fs = require('fs-extra')
 const homePath = require('home-path')
-const mkdir = require('mkdirp')
-const mv = require('mv')
 const npmrc = require('rc')('npm')
 const nugget = require('nugget')
 const os = require('os')
 const path = require('path')
 const pathExists = require('path-exists')
+const semver = require('semver')
+const sumchecker = require('sumchecker')
 
 class ElectronDownloader {
   constructor (opts) {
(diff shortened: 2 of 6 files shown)


kevinsawicki:
@malept this is great, thanks for adding this.

I noticed there was no Windows CI on this repo, so I'm adding AppVeyor over in #26, what do you think about merging that one first and then rebasing this on top? Just to make sure the specs are 🍏 on Windows.

malept:
@kevinsawicki sounds good to me.

test_404.js
```diff
@@ -12,5 +13,4 @@ download({
     throw Error('Download did not throw an error with a custom 404 message: ' + err.message)
   }
   console.log('OK! got expected error:', err.message)
-  process.exit(0)
```
zeke: Why did these exits get taken out?

malept: As far as I can tell, they're not necessary.

kevinsawicki:
#26 has now been merged 🏁

malept:
Excellent. I will try to rebase tonight, and assuming all tests pass on all platforms, I'll merge.

malept:
@fritx I think it's time for a 2.2.0 release.

kevinsawicki:
Quick question, if a file is detected as corrupt, does it still end up in the cache directory?

malept:
@kevinsawicki nope, gets deleted: https://github.com/electron-userland/electron-download/blob/b1ab4ca5d9bc5be95aa0ad76aae5661d876bdd53/index.js#L226-L231

fritx:
@kevinsawicki @malept awesome! :+1:
v2.2.0 has been released.

PR #27, opened 2016-09-10
Fix broken install after ES6 conversion.

danmakenoise:
The usage of `const` outside of the module scope in `index.js` breaks the installation of this package for us, with the following error: 

```SyntaxError: Use of const in strict mode.```

So, I went ahead and fixed it! Here is what I did:

1. Create `eslintrc.json` to specify eslint to check for ES6.
2. Remove use strict inside of modules, where it is not necessary.
3. Revert `index.js` back to using `var` outside of the module scope, where `const` is not allowed.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,0 +1,5 @@
+{
+  "parserOptions": {
+    "ecmaVersion": 6
+  }
+}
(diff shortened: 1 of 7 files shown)


kevinsawicki:
Thanks for this but #28 has been merged which uses babel to precompile the sources files to restore support for node 0.10/0.12 which should address this issue as well.

PR #29, opened 2016-09-12
Run tests via tape

kevinsawicki:
This pull request switches the tests to use [tape](https://github.com/substack/tape) and adds a few more assertions to verify the zip file exists and is non-empty.

This makes them easier to use with other tools that support tap output.

/cc @malept 👀


malept:
:+1: I was not a fan of the existing testsuite.

If the testsuite were larger, I would suggest having a common function that asserted that `error` was null, ZIP path exists and the file is not empty.

kevinsawicki:
> If the testsuite were larger, I would suggest having a common function that asserted that error was null, ZIP path exists and the file is not empty.

Great point, I think that is in scope to this pull request, even though this library is small it makes sense to me to have a very comprehensive test suite given how important this project is, and adding some helpers now will making adding new tests in the future easier with less copy-pasta.

PR #30, opened 2016-09-12
Support old node versions

### LINKED ISSUE #28: Conversion to ES6 should have resulted in a major version change
Due to ES6 conversion, this module no longer works in Node.js environments `< 4`. As a result, while `electron` is perfectly usable in earlier environments, downloading `electron` no longer works. This is causing builds to fail during the `npm install` process.

The workaround here, in order to support earlier environments, would be to either a) revert back to ES5 or b) use `babel` or some other transpiler to transpile the ES6 code to ES5 pre-publish.

kevinsawicki:
This pull requests adds babel support for support old node versions (0.10/0.12).

This precompiles the source files using babel so it does have the downside of not using the native ES6 support that is built into newer versions of node but it does still allow the source files to use these constructs making for cleaner code.

And since the new constructs used, arrow functions and classes, compile down pretty nicely, the resulting babel generated code is still pretty readable.

Refs https://github.com/electron-userland/electron-prebuilt/issues/190
Closes #27
Closes #28 

/cc @malept 👀


malept:
@kevinsawicki I think you're going to have to do something similar to [sumchecker](https://github.com/malept/sumchecker), if you don't mind.

kevinsawicki:
>  I think you're going to have to do something similar to sumchecker, if you don't mind.

Yeah, will do 👍

malept:
Well, that's different: https://travis-ci.org/electron-userland/electron-download/jobs/159396306

kevinsawicki:
> Well, that's different: https://travis-ci.org/electron-userland/electron-download/jobs/159396306

Yeah, these were actually occurring on the builds of #29  as well, https://travis-ci.org/electron-userland/electron-download/jobs/159380783

Super weird, I was chalking it up to a Travis bug/machine issue since I couldn't reproduce locally and the introduction of tape causing these crashes seems unlikely to be related (famous last words).

kevinsawicki:
The crashes on the mac workers were addressed by #32

malept:
> The crashes on the mac workers were addressed by #32

That is bizarre.

malept:
AppVeyor also needs to build against Node 0.10 and 0.12.

package.json
```diff
@@ -25,12 +28,15 @@
     "home-path": "^1.0.1",
     "minimist": "^1.2.0",
     "nugget": "^2.0.0",
-    "path-exists": "^3.0.0",
+    "path-exists": "^2.1.0",
     "rc": "^1.1.2",
     "semver": "^5.3.0",
     "sumchecker": "^1.1.0"
```
malept: IMO sumchecker should be bumped to enforce a minimum version.

malept: So weird to get a notification for my own commit...

kevinsawicki:
Thanks @malept for your help, review, and sumchecker updates on this one 👍

PR #37, opened 2016-10-15
Save cache to standard directory

Siilwyn:
The cache is saved to a standard cache directory using the "env-paths"
module, rather than saving in the home directory.
The old cache directory will automatically be migrated to the new
location.

BREAKING CHANGE: Cache is no longer stored in `~/.electron`.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -1,6 +1,7 @@
 'use strict'
 
 const debug = require('debug')('electron-download')
+const envPaths = require('env-paths')
 const fs = require('fs-extra')
 const homePath = require('home-path')
 const rc = require('rc')
@@ -44,7 +45,8 @@ class ElectronDownloader {
   }
 
   get cache () {
-    return this.opts.cache || path.join(homePath(), './.electron')
+    // use passed argument or XDG environment variable fallback to OS default
+    return this.opts.cache || envPaths('electron-download', {suffix: ''}).cache
   }
 
   get cachedChecksum () {
@@ -171,18 +173,25 @@ class ElectronDownloader {
   }
 
   createCacheDir (cb) {
-    fs.mkdirs(this.cache, (err) => {
-      if (err) {
-        if (err.code !== 'EACCES') return cb(err)
-        // try local folder if homedir is off limits (e.g. some linuxes return '/' as homedir)
-        let localCache = path.resolve('./.electron')
-        return fs.mkdirs(localCache, function (err) {
-          if (err) return cb(err)
-          cb(null, localCache)
-        })
-      }
-      cb(null, this.cache)
-    })
+    try {
+      // if files exist in the old path use and migrate it
+      const oldPath = path.join(homePath(), './.electron')
+      fs.renameSync(oldPath, envPaths('electron-download', {suffix: ''}).cache)
+    } catch (e) {
+      fs.mkdirs(this.cache, (err) => {
+        if (err) {
+          if (err.code !== 'EACCES') return cb(err)
+
+          // try local folder if homedir is off limits (e.g. some linuxes return '/' as homedir)
+          let localCache = path.resolve('./.electron')
+          return fs.mkdirs(localCache, function (err) {
+            if (err) return cb(err)
+            cb(null, localCache)
+          })
+        }
+      })
+    }
+    cb(null, this.cache)
   }
 
   createTempDir (cb, onSuccess) {
--- a/lib/index.js
+++ b/lib/index.js
@@ -46,7 +46,7 @@ class ElectronDownloader {
 
   get cache () {
     // use passed argument or XDG environment variable fallback to OS default
-    return this.opts.cache || envPaths('electron-download', {suffix: ''}).cache
+    return this.opts.cache || envPaths('electron', {suffix: ''}).cache
   }
 
   get cachedChecksum () {
@@ -176,7 +176,7 @@ class ElectronDownloader {
     try {
       // if files exist in the old path use and migrate it
       const oldPath = path.join(homePath(), './.electron')
-      fs.renameSync(oldPath, envPaths('electron-download', {suffix: ''}).cache)
+      fs.renameSync(oldPath, envPaths('electron', {suffix: ''}).cache)
     } catch (e) {
       fs.mkdirs(this.cache, (err) => {
         if (err) {
--- a/lib/index.js
+++ b/lib/index.js
@@ -53,6 +53,11 @@ class ElectronDownloader {
   }
 
   get cachedZip () {
+    const oldLocation = path.join(os.homedir(), './.electron', this.filename)
+    if (pathExists.sync(oldLocation)) {
+      return oldLocation
+    }
+
     return path.join(this.cache, this.filename)
   }
 
--- a/lib/index.js
+++ b/lib/index.js
@@ -178,25 +178,18 @@ class ElectronDownloader {
   }
 
   createCacheDir (cb) {
-    try {
-      // if files exist in the old path use and migrate it
-      const oldPath = path.join(homePath(), './.electron')
-      fs.renameSync(oldPath, envPaths('electron', {suffix: ''}).cache)
-    } catch (e) {
-      fs.mkdirs(this.cache, (err) => {
-        if (err) {
-          if (err.code !== 'EACCES') return cb(err)
-
-          // try local folder if homedir is off limits (e.g. some linuxes return '/' as homedir)
-          let localCache = path.resolve('./.electron')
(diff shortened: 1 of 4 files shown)


Siilwyn:
The failing tests do not seem related to this but have to do with older Node.js versions. Maybe bundle this change with removing 0.10 and 0.12 support and then do a major version bump?

malept:
> Maybe bundle this change with removing 0.10 and 0.12 support and then do a major version bump?

Removing Node 0.12 support won't happen until LTS is over for that series. I learned that the hard way.

malept:
I don't suppose there is an `env-paths` version (or similar module) that has Node 0.12 support?

Siilwyn:
Unfortunately not, it's ES6 from the start. :*)
I can make a compatible one very easily or we have to wait for next year, your call...

malept:
Personally, I'm OK with waiting. In the meantime though, is it possible to add tests, particularly for the cache directory migration behavior?

Siilwyn:
Great idea, I discovered my implementation only to be working partially! Now it's all good, added a test. :bulb:

lib/index.js
```diff
@@ -37,7 +38,8 @@ class ElectronDownloader {
   }
 
   get cache () {
-    return this.opts.cache || path.join(homePath(), './.electron')
+    // use passed argument or XDG environment variable fallback to OS default
+    return this.opts.cache || envPaths('electron-download', {suffix: ''}).cache
```
malept: I wonder if it should be `electron` rather than `electron-download`.

test/test.js
```diff
@@ -15,3 +17,27 @@ test('Basic test', (t) => {
     t.end()
   })
 })
+
+test('Cache directory is moved to new location', (t) => {
+  // Download to old cache directory location
```
malept: I don't know whether this is the right way to write this test. Perhaps if the user specifies a custom cache directory, the migration code shouldn't be triggered.

Regardless of the decision reached for the above question, I think a better test would be:

1. Create `$HOME/.electron` if it doesn't exist, and put a temporary file in that folder
2. Invoke `download()`
3. Assert that the temporary file exists in the new default cache directory
4. Delete temporary file

Siilwyn: :thinking: It's only temporary code though just for migrating, after some time it will be removed... Already spent a lot of time on it, apart from cleaning up the file it does the same as the steps you proposed right?

Thoughts on how important this is? Not saying I don't want to do it at all but I'm just considering the time it costs versus value that comes out of it.

malept: I don't know about "temporary". I think the migration code is going to stick around for at least six months after release, since not everyone immediately upgrades.

Siilwyn: So ...? Apart from cleaning up the file it does the same as the steps you proposed right?

malept: I would prefer it if it was more explicit.

Siilwyn: Alright will rewrite it a bit then.

lib/index.js
Siilwyn: Not sure since this is for the module `electron-download` thought it would make sense to name it like that... But since it's for `electron` it might make sense to name it like that.

malept: Yeah, ultimately the `electron` module uses this. Now that you've reminded me, it's worth documenting where the default cache directory is on Windows/macOS/Linux in the README.

Siilwyn: Okay, added locations to the readme and renamed it to 'electron'.

malept:
FWIW, [`cachedir`](https://www.npmjs.com/package/cachedir) provides the same functionality (as of today :smile:) and works on Node < 4.

Siilwyn:
Think `env-paths` is a better fit because it's used in more modules which has multiple advantages. Also a module to only return a path to the cache directory is a bit too minimal for my taste. :) Thanks for the heads-up though! 2016-12-31 is almost there! :-)

thesbros:
Just checking, is this scheduled to be merged once 0.12 is no longer supported (end of the month)?

malept:
That's my plan, unless one of the other maintainers has concerns.

test/test.js
```diff
@@ -15,3 +19,31 @@ test('Basic test', (t) => {
     t.end()
   })
 })
+
+test.only('Cache directory is moved to new location', (t) => {
```
kevinsawicki: This `.only` call should be removed.

Siilwyn: Whoops, fixed. :D

Siilwyn:
11 days since 0.12 got deprecated. :rocket:

malept:
@Siilwyn There seems to be a merge conflict.

_(If you're wondering why this hasn't gotten merged yet, aside from the merge conflict, the Electron maintainers are taking removal of Node version support very seriously, because this affects the `electron` module on NPM, and a fair amount of discussion has been generated. We'll know more next week.)_

thesbros:
Another alternative is using a module compatible with Node 0.12, if the maintainers don't feel comfortable removing support.

malept:
@thesbros I mentioned a 0.12-compatible module in https://github.com/electron-userland/electron-download/pull/37#issuecomment-266480318, with the response right below it.

Siilwyn:
Hey also tuning in. :)
Let's await next week's outcome. I hope they choose to move forward and leave 0.12 behind us, I also [suggested writing](https://github.com/electron-userland/electron-download/pull/37#issuecomment-261184058) a 0.12 compatible module earlier. However I think it's better to depend on a wildly used module.

Siilwyn:
Ugh, just tried rebasing, every commit gives a different merge conflict.
Will look into it later again, reverted the rebase for now.

Siilwyn:
Not sure if I did or not but it seems there are no conflicts with master anymore. :)

malept:
Maintainer update: discussion is still ongoing, I expect to have more information next week.

Siilwyn:
Alright, thanks for the quick reply @malept !

Siilwyn:
Any update, can I read the discussion anywhere or is it verbal?

malept:
@Siilwyn 
We've agreed to move to drop Node < 4 support (an official announcement will hopefully be in the near future), and so I've opened #45 as a result.

During that discussion, we discussed this PR in particular. Some of the maintainers have reservations about accepting this from a conceptual point of view. @kevinsawicki or @zeke, could one of you chime in so I don't misconstrue what was said during the meeting?

Siilwyn:
:(

Siilwyn:
Any update on this?

malept:
@kevinsawicki ping

Siilwyn:
@kevinsawicki could you please review this again?

kevinsawicki:
> could you please review this again?

I'm not a huge fan of moving a directory each time this runs. Other apps or libraries could be using `~/.electron` and it would seem confusing and hard to track down why this directory is disappearing each time you `npm install electron`.

Could we instead just use `~/.electron` as a fallback location if the file isn't in the new cache location and write any new files to the new location?

That way it would be backwards compatible still but we wouldn't have to worry about moving directories around.

kevinsawicki:
It also looks like there are conflicts here as well that need to be resolved.

Siilwyn:
Will remove the migration in that case.

I don't want to be pointing fingers and get negative but would really like to get this off my chest: the migration of the directory was suggested to me and I spent a lot of time getting that working and covered by a test. Solving the merge conflicts is something that happened multiple times now on this PR because of the slow responses.

zeke:
Sorry to leave this hanging for so long @Siilwyn, and thank you for patiently rebasing and making suggested changes.

I see one failure on Travis:

```
not ok 17 Zip path should end with -dsym.zip
  ---
    operator: ok
    expected: true
    actual:   false
    at: download (/Users/travis/build/electron-userland/electron-download/test/test_dsym.js:16:7)
```

Siilwyn:
Thank you for your reply Zeke, I just added some fallback code all tests are passing now!

readme.md
```diff
@@ -25,7 +25,7 @@ download({
   version: '0.25.1',
   arch: 'ia32',
   platform: 'win32',
-  cache: './zips' // defaults to <user's home directory>/.electron
+  cache: './zips' // defaults to <user's cache directory>/electron-download
```
kevinsawicki: Should this be `defaults to <user's cache directory>/electron` instead?

Siilwyn: Yes, sharp eye!
This is a leftover from before the 'Cache location' section got added so I removed it as it's duplicate.

kevinsawicki:
Thanks for this @Siilwyn, apologies for the mixed signals, and appreciate the resiliency 👍 🚀

Siilwyn:
Nice to see this getting merged! :)

PR #40, opened 2016-11-18
bump dependencies, require node 4, remove home-path dependency

develar:
electron-builder, electron-packager and electron-osx-sign required node 4, so, we can 
* reduce number of dependencies and use `os.homedir()`,
* use babel preset `es2015-node4`.

Also, dependencies updated — fs-extra updated to 1.0.0.

`path-exists` updated to 3.0.0 — node <4 support is dropped (https://github.com/sindresorhus/path-exists/pull/4).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,11 +5,10 @@ os:
 osx_image: xcode7.3
 language: node_js
 node_js:
-- '0.10'
-- '0.12'
 - '4'
 - '5'
 - '6'
+- '7'
 cache:
 - directories:
   - "$HOME/.npm"
--- a/lib/index.js
+++ b/lib/index.js
@@ -2,7 +2,6 @@
 
 const debug = require('debug')('electron-download')
 const fs = require('fs-extra')
-const homePath = require('home-path')
 const rc = require('rc')
 const nugget = require('nugget')
 const os = require('os')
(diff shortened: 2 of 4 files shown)


malept:
I already tried that once (see https://github.com/electron-userland/electron-download/pull/30). We can't drop Node < 4.0 support until 2017, because `electron-prebuilt` depends on `electron-download`, and Node 0.12 is officially supported until the end of 2016. My plan was to remove babel from `electron-download` (and upgrade `path-exists`) after January 1.

mafintosh:
There is a big difference between this module and the signing stuff. This one is used when people install electron (for example when people use https://github.com/mappum/electron-webrtc in node apps). I don't see the need to drop support for any node unless we *really* have to.

malept:
I think it's reasonable to drop any Node version that the NodeJS team no longer maintains.

wtgtybhertgeghgtwtg:
One argument for dropping support for older versions is that it would allow `electron-download` to use newer dependencies.  For example, you could replace `nugget` or `request` with `got` and significantly reduce the dependency tree.

develar:
@wtgtybhertgeghgtwtg it is what should be done, but nugget also provides progress bar — got doesn't provide this functionality. Task is not to easy.

wtgtybhertgeghgtwtg:
If you can get `nugget` to bump their minimum required `node` version to at least 4, it can be changed to use `got` over `request`.

develar:
@malept Conflicts resolved. Today is 1 Feb 2017 — time to remove < node 4 support :)

PR #43, opened 2016-12-27
Add force option to redownload

### LINKED ISSUE #38: Clear or ignore cache
It'd be really helpful if I could either clear the cache via the CLI, or instruct `electron-download` to ignore cache during install via the API & CLI.

I have a workflow for a project, which involves republishing custom electron binaries without changing the name or version. I know, [it's a little bit like this xkcd](https://xkcd.com/1172/), but it saves me some headache if I can republish and just clear the cache automatically. (It lets me avoid making lots of commits that change a version number during development.)

A command which clears the cache, or a way to instruct the downloader to overwrite the cache, would be fine. Happy to make a PR based on your preference.

kevinsawicki:
Add a `force` boolean option that when set will re-download the asset and the `SHASUM` file to the cache area.

/cc @pfrazee would this option work for your use case?

Closes #38

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -105,6 +105,10 @@ class ElectronDownloader {
     return strictSSL
   }
 
+  get force () {
+    return this.opts.force || false
+  }
+
   get symbols () {
     return this.opts.symbols || false
   }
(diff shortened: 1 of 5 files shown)


pfrazee:
Yes definitely. Thank you!

PR #45, opened 2017-02-02
Remove Node 0.12 support and Babel

malept:
Hopefully this goes a bit better than when I removed both of these things from sumchecker.

Per the policy, this will require a major version bump.

Closes #40.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,3 @@
 node_modules
 .DS_Store
-build
 npm-debug.log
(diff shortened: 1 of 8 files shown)


malept:
> I'm going to merge #46 first though and cut a minor release before merging this and bumping the major.

:+1: sounds good.

kevinsawicki:
🚢 🚢 🚢 🚢 published as 4.0.0, thanks again 👍

Issue #24, opened 2016-08-30
Save cached releases to standard directories


Siilwyn:
Instead of placing the release archives in the home directory they should be placed in the appropriate directory per OS. I've done the same for [Greenkeeper's RC](https://github.com/greenkeeperio/rc/pull/3) so I'd like to create a PR for it.

Before doing that I'd like to know if there are any special things I should take into consideration and if the core contributors are on board with this so I don't do it in vain. :)

malept:
What do you propose to do with the existing cache directory? Is that still taken into account?

Siilwyn:
We could use the old location if it exists or move the old location to the new location. At Greenkeeper the moving is done. Both have their own pros and cons.

Siilwyn:
@malept any preference for one of the solutions?

malept:
I would prefer moving it.

develar:
@Siilwyn I see error in your merged changes, that leads to https://github.com/electron-userland/electron-download/pull/47#issuecomment-292803245

Code
```
const oldLocation = path.join(os.homedir(), './.electron', this.filename)
    if (pathExists.sync(oldLocation)) {
      return oldLocation
    }
```

in the `cachedZip`

But! We call `cachedZip` in the `ensureCacheDir` twice — one for check, and another one in the call `downloadZip`. So, 2 sync calls are performed, and latter can lead to race conditions — `downloadZip` must  download only to a new location.

In version of electron-download that used by electron-builder, I will remove this code for now.

Siilwyn:
Oh that is bad. :/

@develar could you expand your explanation, I don't quite follow.

develar:
@Siilwyn downloadZip calls cachedZip, but cachedZip will check and return old location. Not evident how does it lead to error, but in any case it is not good — cachedZip should return constant result (based on options) and old location should be checked in another place.

Siilwyn:
The function that checks if the zip file is already cached sounds like the appropriate place for this check to me. Where would you place it?

Siilwyn:
@develar [moving the logic to the directory location](https://github.com/electron-userland/electron-download/compare/master...Siilwyn:fix-fallback-location) seems to fix it. Could you confirm this?

PR #47, opened 2017-02-24
Atomic rename

develar:
https://github.com/electron-userland/electron-download/pull/36#issuecomment-261473619

Already used in the electron-builder. During test period no issues were discovered.

Close #36

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -76,6 +76,8 @@ class ElectronDownloader {
       return `ffmpeg-${suffix}.zip`
     } else if (this.symbols) {
       return `electron-${suffix}-symbols.zip`
+    } else if (this.dsym) {
+      return `electron-${suffix}-dsym.zip`
     } else {
       return `electron-${suffix}.zip`
     }
@@ -112,6 +114,10 @@ class ElectronDownloader {
 
   get symbols () {
     return this.opts.symbols || false
+  }
+
+  get dsym () {
+    return this.opts.dsym || false
   }
 
   get chromedriver () {
--- a/lib/index.js
+++ b/lib/index.js
@@ -10,6 +10,8 @@ const pathExists = require('path-exists')
 const semver = require('semver')
 const sumchecker = require('sumchecker')
 
+let tmpFileCounter = 0
+
 class ElectronDownloader {
   constructor (opts) {
     this.opts = opts
@@ -140,12 +142,8 @@ class ElectronDownloader {
     pathExists(this.cachedChecksum).then(exists => {
       if (exists && !this.force) {
         this.verifyChecksum(cb)
-      } else if (this.tmpdir) {
-        this.downloadChecksum(cb)
       } else {
-        this.createTempDir(cb, (callback) => {
-          this.downloadChecksum(callback)
-        })
+        this.downloadChecksum(cb)
       }
     })
   }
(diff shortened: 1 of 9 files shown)


develar:
@malept Could you please review?

MarshallOfSound:
Probably want a second set of eyes before merge into a key module

/cc @electron-userland

lib/index.js
```diff
@@ -190,23 +188,16 @@ class ElectronDownloader {
...
-
   downloadChecksum (cb) {
-    this.downloadFile(this.checksumUrl, this.checksumFilename, this.cachedChecksum, cb, this.verifyChecksum.bind(this))
+    this.downloadFile(this.checksumUrl, this.cachedChecksum, cb, this.verifyChecksum.bind(this))
   }
 
-  downloadFile (url, filename, cacheFilename, cb, onSuccess) {
-    debug('downloading', url, 'to', this.tmpdir)
+  downloadFile (url, cacheFilename, cb, onSuccess) {
+    const tempFileName = `tmp-${process.pid}-${(tmpFileCounter++).toString(16)}-${path.basename(cacheFilename)}`
```
malept: I'd rather have this use something like [tempnam](https://www.npmjs.com/package/tempnam) instead of introducing global variables and hand-rolling it.

develar: `tempnam`  is not suitable because using fs. I think it is ok to not use external module for two lines of code (also, to reduce number of external deps and increase security).

lib/index.js
```diff
@@ -10,6 +10,8 @@ const pathExists = require('path-exists')
 const semver = require('semver')
 const sumchecker = require('sumchecker')
 
+let tmpFileCounter = 0
```
malept: If you're going to keep this, at least make it a class variable or something.

lib/index.js
malept: > `tempnam` is not suitable because using fs.

Could you please explain this?

lib/index.js
```diff
@@ -253,13 +244,18 @@ class ElectronDownloader {
...
-      fs.move(path.join(this.tmpdir, filename), target, (err) => {
-        if (err) return cb(err)
-        onSuccess(cb)
-      })
+    const cache = this.cache
+    debug('moving', filename, 'from', cache, 'to', target)
+    fs.rename(path.join(cache, filename), target, (err) => {
+      if (err) {
+        try {
+          fs.unlinkSync(cache)
```
malept: Why not use the async version and get rid of the try...catch?

lib/index.js
```diff
@@ -253,13 +244,18 @@ class ElectronDownloader {
...
+        } catch (error) {
+          console.error(`Error deleting cache file: ${error.message}`)
```
malept: This needs to show the filename that couldn't be removed so the user can deal with it themselves.

lib/index.js
develar: I don't want to use something like [mkdtemp](https://nodejs.org/api/fs.html#fs_fs_mkdtemp_prefix_options_callback) here because only our tool creates temp file here — `process.pid` allows us to be sure that will be no conflicts with other processes. To not complicate and keep it simple.

lib/index.js
develar: It must be file level variable because class can be created several times.

develar:
I have answered to all questions, what's left?

kevinsawicki:
> I have answered to all questions, what's left?

This looks good to me, what do you think @malept?

lib/index.js
malept: Yes, a class variable, not an instance variable.

lib/index.js
```diff
@@ -253,13 +244,22 @@ class ElectronDownloader {
...
+    fs.rename(path.join(cache, filename), target, (err) => {
+      if (err) {
+        fs.unlink(cache, cleanupError => {
+          try {
+            if (cleanupError) {
+              console.error(`Error deleting cache file: ${cleanupError.message}`)
+            }
+          } finally {
+            cb(err)
+          }
```
malept: Not sure what the point of the `try...finally` is. Is an exception being thrown?

lib/index.js
malept: I don't think this was addressed.

malept:
@zeke given #36, do you want to take a look?

lib/index.js
zeke: ☝️  Yeah it seems unnecessary...

develar: User application can override `console.error`. e.g. Jest does it. As we must call user `cb` in any case`, it is more safe to use `try-catch` here. Otherwise we can get unresolved promise and so on.

zeke: > Jest does it

Yikes. Ok.

lib/index.js
develar: > a class variable, not an instance variable.

Hmm... ES6 doesn't support 

```
static tmpFileCounter = 0
```

> SyntaxError: Unexpected token =

We cannot use this syntax because node doesn't support it.

What do you mean? And for what we should use class variable here? It should be module level variable. e.g. I write in Kotlin and prefer to use file level vars instead of static.

MarshallOfSound: @develar That syntax isn't supported but the syntax `ClassName.tmpFileCounter = 0` does work and has the same affect 👍

develar:
I don't like `ElectronDownloader.tmpFileCounter` (local module variable is better and more safe), but if you want — ok.

develar:
Right now electron-download should be not released after #24 — I get strange errors like `Error: ENOENT: no such file or directory, open '/Users/develar/Library/Caches/electron/electron-v1.6.3-darwin-x64.zip' And since electron-download (and used libs) doesn't use promises with long stack traces, it is not easy to find cause of error. Not yet clear — is it due to my changes or upstream affected as well.

Siilwyn:
@develar I'm assuming you already found the culprit pointed out in #24

develar:
Friendly ping :)

kevinsawicki:
Thanks @develar 👍 🚢

PR #50, opened 2017-04-04
allow redirects

Zougi:
fix issue https://github.com/electron-userland/electron-download/issues/49


Zougi:
When I download electron-download project master branch, do npm install && npm test, npm test fails the same way...maybe it's because I'm behind a proxy... HTTP_PROXY and HTTPS_PROXY variable are defined however...

PR #53, opened 2017-04-08
mirror that works

Zougi:
- **fix follow redirect**
I replaced nugget by request which is more robust. nugget considers all http response code above 299 to be errors...it's not the case!!
fix https://github.com/electron-userland/electron-download/issues/49
- **allow mirror to not host a checksum**
at work I really cannot host on our common repository the sha. I cannot download electron's sha neither from an outside url
fix https://github.com/electron-userland/electron-download/issues/51

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -4,7 +4,7 @@ const debug = require('debug')('electron-download')
 const envPaths = require('env-paths')
 const fs = require('fs-extra')
 const rc = require('rc')
-const nugget = require('nugget')
+const request = require('request')
 const os = require('os')
 const path = require('path')
 const pathExists = require('path-exists')
@@ -142,7 +142,10 @@ class ElectronDownloader {
   }
 
   get verifyChecksumNeeded () {
-    return semver.gte(this.version, '1.3.2')
+    return !(this.opts.electron_no_checksum === true ||
+      this.npmrc.electron_no_checksum === 'true' ||
+      process.env.ELECTRON_NO_CHECKSUM === 'true' ||
+      semver.lt(this.version, '1.3.2'))
   }
 
   get version () {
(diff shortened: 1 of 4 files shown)


MarshallOfSound:
@Zougi Personally I quite like nugget (the progress bar it gives is pretty good for slower network connections).  I would prefer you fixed the `>299`  error code issue in nugget rather than replacing the dependency

Zougi:
='(

Zougi:
I could add a progress bar with https://github.com/visionmedia/node-progress/ or https://www.npmjs.com/package/request-progress but I won't fix nugget

Zougi:
@MarshallOfSound I don't have the pretention to know enough about http(s) specificities in order to fix the package. However here is a progressbar

Zougi:
Hi, I know @kevinsawicki @malept , I know you're busy on other electron projects, and thanks for your continuous contributions, I draw your attention to this PR, could you please take the time to review it? It sounds silly but without this PR I cannot use electron in the company I work for. Indeed, we have a very strict build process and are not allowed to make calls to a network outside the company during the build of our softwares. Electron must therefore be hosted on an internal mirror. The repository we hosts our code and third parties sends a 302 redirect which is not supported **yet** by electron-download

malept:
Replacing the HTTP stack is quite a big change. I agree with @MarshallOfSound, I would rather have the fix in nugget.

TekSiDoT:
Having a hard coded checksum download is breaking all builds that require electron-packager to be used in an strictly offline environment. I honestly dont get why skipping an external checksum download as an option warrants this discussion.

sla89:
@Zougi in my company we do not allow to make calls to a network outside the company's. We solved this by setting electron_mirror in our `.yarnrc`. This also works for `npm`.

sla89:
@malept, @MarshallOfSound, @kevinsawicki did you know that `nugget` seems to be dead? One of the contributes said 14 days ago that he has no time to maintain it (see maxogden/nugget#27).

MarshallOfSound:
Closing as `next` completely changes the way downloading works, powered by got, customizable, etc.

PR #54, opened 2017-04-14
Fix fallback location

Siilwyn:
See @develar's concerns at the original issue #24 and at the [atomic rename PR](https://github.com/electron-userland/electron-download/pull/47#issuecomment-292803245).

I have included a test that fails against master and is passing by moving the fallback logic to directory level.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -44,13 +44,14 @@ class ElectronDownloader {
   }
 
   get cache () {
+    if (this.opts.cache) return this.opts.cache
+
     const oldCacheDirectory = path.join(os.homedir(), './.electron')
     if (pathExists.sync(path.join(oldCacheDirectory, this.filename))) {
-      return path.join(os.homedir(), './.electron')
+      return oldCacheDirectory
     }
-
     // use passed argument or XDG environment variable fallback to OS default
-    return this.opts.cache || envPaths('electron', {suffix: ''}).cache
+    return envPaths('electron', {suffix: ''}).cache
   }
 
   get cachedChecksum () {
--- a/lib/index.js
+++ b/lib/index.js
@@ -44,6 +44,11 @@ class ElectronDownloader {
   }
 
   get cache () {
+    const oldCacheDirectory = path.join(os.homedir(), './.electron')
+    if (pathExists.sync(path.join(oldCacheDirectory, this.filename))) {
+      return path.join(os.homedir(), './.electron')
+    }
+
     // use passed argument or XDG environment variable fallback to OS default
     return this.opts.cache || envPaths('electron', {suffix: ''}).cache
   }
@@ -53,11 +58,6 @@ class ElectronDownloader {
   }
 
   get cachedZip () {
-    const oldLocation = path.join(os.homedir(), './.electron', this.filename)
-    if (pathExists.sync(oldLocation)) {
-      return oldLocation
-    }
-
     return path.join(this.cache, this.filename)
   }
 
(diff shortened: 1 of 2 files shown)


Siilwyn:
So my PR introduced a :bug: that causes an error when trying to download a new version that is cached in the original location.
Just pushed again, CI was failing because of the linter, tagging those involved in the previous PR... @MarshallOfSound @zeke @kevinsawicki

develar:
Are we really need to worry about old cache?

lib/index.js
```diff
@@ -44,6 +44,11 @@ class ElectronDownloader {
   }
 
   get cache () {
+    const oldCacheDirectory = path.join(os.homedir(), './.electron')
+    if (pathExists.sync(path.join(oldCacheDirectory, this.filename))) {
```
develar: I think, we should avoid `sync`.

Siilwyn: Async code is often better but I chose for a sync method in this case. Firstly because the path to the cache is needed for all other functionality. Secondly all other code expects this method to be sync. Lastly this fallback will be removed in the future so rewriting other methods to work in an async way with `get cache` sounds unreasonable.

Siilwyn:
@develar see the original PR and in particular [this comment](https://github.com/electron-userland/electron-download/pull/37#issuecomment-289506600) on why we worry about the old cache.

Siilwyn:
Sorry I forgot to tag @malept who was also involved in the previous PR.
By the way could somebody re-run CI so it's green, thanks!

Siilwyn:
@kevinsawicki thoughts?

lib/index.js
```diff
@@ -44,6 +44,11 @@ class ElectronDownloader {
...
+      return path.join(os.homedir(), './.electron')
+    }
+
     // use passed argument or XDG environment variable fallback to OS default
     return this.opts.cache || envPaths('electron', {suffix: ''}).cache
```
kevinsawicki: Shouldn't this method always return `this.opts.cache` when specified? Even if the old directory exists.

test/test_fallback.js
```diff
@@ -0,0 +1,33 @@
...
+    arch: 'x64',
+    platform: 'linux',
+    quiet: true
+  }
+
+  download(
+    Object.assign(
+      {},
+      downloadOptions,
+      { cache: oldCachePath }
```
kevinsawicki: Is there a reason for the three objects here instead of just setting `cache` directly on `downloadOptions`?

lib/index.js
Siilwyn: Sharp! Added a test to confirm this is working. Also found the first test could be improved.

test/test_fallback.js
Siilwyn: This is because I re-use the same settings object with the only difference being the `cache` property. It looks a bit wonky because of the newlines though, removed those.

lib/index.js
```diff
@@ -44,20 +44,21 @@ class ElectronDownloader {
   }
 
   get cache () {
+    if (this.opts.cache) return this.opts.cache
+
+    const oldCacheDirectory = path.join(os.homedir(), './.electron')
+    if (pathExists.sync(path.join(oldCacheDirectory, this.filename))) {
+      return path.join(os.homedir(), './.electron')
```
kevinsawicki: This could just return `oldCacheDirectory` I believe.

kevinsawicki:
Thanks @Siilwyn for fixing this 👍 🚢

Siilwyn:
Thanks for your helpful reviews too @kevinsawicki! :tada:

PR #55, opened 2017-05-10
Remove fallback location and document how to migrate

kevinsawicki:
Instead of supporting a fallback location to the old cache directory, this pull request removes it and just documents how people can move it over.

It seems worthwhile to just do a clean break here to simplify the code and have consistent behavior in terms of where assets are pull from.

Refs #54 
Refs #37

/cc @malept @Siilwyn

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -53,11 +53,6 @@ class ElectronDownloader {
   }
 
   get cachedZip () {
-    const oldLocation = path.join(os.homedir(), './.electron', this.filename)
-    if (pathExists.sync(oldLocation)) {
-      return oldLocation
-    }
-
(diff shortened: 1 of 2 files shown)


kevinsawicki:
Thought about this some more, I think it makes sense to go with the fallback location to start since to avoid bumping the major, can revisit in the future though possibly when other major API changes are also in the works.

Siilwyn:
Phew, so I didn't waste those hours after all. :stuck_out_tongue: 
I actually don't like fallbacks so much but for much depended on modules it might make sense. Benefit is also that after a while the old cache will only contain old electron versions which are probably used less in the future.

PR #56, opened 2017-05-10
Support new version in Chromedriver asset name

kevinsawicki:
Support new version in asset name starting in 1.7.0

Refs https://github.com/electron/electron/pull/8682

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -75,7 +75,12 @@ class ElectronDownloader {
     const suffix = `v${this.version}-${type}`
 
     if (this.chromedriver) {
-      return `chromedriver-v2.21-${type}.zip`
+      // Chromedriver started using Electron's version in asset name in 1.7.0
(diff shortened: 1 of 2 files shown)


kevinsawicki:
> I assume this needs to be backported?

Do you mean to electron-download@1.3.x line?

malept:
> > I assume this needs to be backported?
>
> Do you mean to electron-download 3.x line?

Yeah, I meant to the version that electron-prebuilt uses.

kevinsawicki:
> Yeah, I meant to the version that electron-prebuilt uses.

I could definitely do that, although this won't impact `prebuilt` since it does not download Chromedriver, the only consumer of the chromedriver download path that I know of is https://github.com/electron/chromedriver which I was already going to upgrade to 4.x

malept:
Ah, OK.  Sounds like that's not necessary, then.

PR #59, opened 2017-07-12
chore: update fs-extra to 4

develar:
To reduce client node_modules size.

note: I am going to send number of PR and so, when this PR will be accepted, you can postpone publish to npm until all PR will be not accepted.


MarshallOfSound:
Totally didn't show up in my notifications till just now 😕 

Overally looks fair enough, looked at the changelogs for all bumped modules and it looks OK.  One concern it the adoption of `yarn`.  No other core repositories (`electron`, `electron-prebuilt`, etc.) have adopted yarn yet and in general lock files don't work well cross-platform (so electron dev with yarn is kinda not working too well).  This might just need to be a discussion we have somewhere that says "should we use it".  But for now, if we remove it from this PR I'm happy to approve it.

develar:
I just want to speedup installation on CI (according to my experience, yarn works better than npm without global cache (cache on CI doesn't make any sense)). But of course — it depends and for this project doesn't matter. Removed.

>  didn't show up in my notifications till just now

Yeach... Still no luck to get rid of electron-download-tf because when one my PR is accepted, I need to apply another one :(

package.json
```diff
@@ -22,23 +22,24 @@
...
   "homepage": "https://github.com/electron-userland/electron-download#readme",
   "dependencies": {
-    "debug": "^2.2.0",
+    "debug": "^3.0.0",
     "env-paths": "^1.0.0",
-    "fs-extra": "^2.0.0",
+    "fs-extra": "^4.0.1",
     "minimist": "^1.2.0",
-    "nugget": "^2.0.0",
+    "nugget": "^2.0.1",
```
malept: Unnecessary noise

package.json
```diff
@@ -22,23 +22,24 @@
...
     "path-exists": "^3.0.0",
-    "rc": "^1.1.2",
-    "semver": "^5.3.0",
-    "sumchecker": "^2.0.1"
+    "rc": "^1.2.1",
+    "semver": "^5.4.1",
+    "sumchecker": "^2.0.2"
```
malept: Unnecessary noise

develar:
@malept Yarn is used only on CI server and doesn't impact local users. yarn.lock removed.

package.json
develar: I think it worth to explicitly note that tests and manual testing were performed with at least these versions.

malept: I still don't think it's worth bumping unless it's necessary for a new feature/bugfix.

develar: > unless it's necessary for a new feature/bugfix.

Because `^` is used here, it means that users will get any new versions. I like ideas described in the https://yarnpkg.com/blog/2016/11/24/lockfiles-for-all/ 

I believe that it is better to specify new version and more recent if possible. It doesn't harm at all. It is only one changed symbols (Of course, no one insist to publish new module version only because some dependency has a new minor version.) When module will be installed — package tool in any case will  resolve to latest allowed version.

As you can see, first commit in this PR was failed due to NPM bug (cache). That's why yarn is used on CI (reliable and fast) and I prefer to always just use `npm-check-updates -a` (and read changelings, of course).

MarshallOfSound:
@develar Only thing blocking this rn is the yarn stuff which we don't want to adopt here without a broader discussion, it might make tests slightly faster but it also results in different testing conditions to those people using npm.

A wider discussion needs to be had about supporting `yarn` with `-userland` modules, in particular things like cross-platform lock files need to be talked about.  But this PR isn't the place for that, happy to merge this once yarn is moved out and we can open a discussion issue somewhere else RE yarn.

develar:
@MarshallOfSound Thanks a lot for breaking the silence. `yarn` removed on CI servers.

MarshallOfSound:
@develar Second time the build timed out on macOS, I reran it the first time.  Any ideas whether the `fs` bump could have caused the hang?

https://travis-ci.org/electron-userland/electron-download/jobs/270780729

develar:
@MarshallOfSound It was ok 16 days ago (and when PR was submitted). I will check.

develar:
Stability — electron-builder uses electron-download-tf (this PR and #60 and not yet filed muon support). So, it is widely tested already.

PR #61, opened 2017-08-02
get baseUrl, middleUrl and filename from package

benlancaster:
There were some inconsistencies here; you could specify the mirror base URL in npm config and .npmrc, but not middleUrl or urlSuffix (filename). This change rectifies that and allows you to specify baseUrl, middleUrl and urlSuffix in the consuming project's [package.json `config` block](https://docs.npmjs.com/files/package.json#config)

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -24,28 +24,28 @@ class ElectronDownloader {
   }
 
   get baseUrl () {
-    return process.env.ELECTRON_MIRROR ||
-      process.env.npm_package_config_electron_mirror ||
+    return process.env.NPM_CONFIG_ELECTRON_MIRROR ||
       process.env.npm_config_electron_mirror ||
-      process.env.NPM_CONFIG_ELECTRON_MIRROR ||
+      process.env.npm_package_config_electron_mirror ||
+      process.env.ELECTRON_MIRROR ||
       this.opts.mirror ||
       'https://github.com/electron/electron/releases/download/v'
   }
 
   get middleUrl () {
-    return process.env.ELECTRON_CUSTOM_DIR ||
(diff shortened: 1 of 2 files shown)


lib/index.js
```diff
@@ -24,19 +24,30 @@ class ElectronDownloader {
...
+      this.opts.customDir ||
+      this.version
   }
 
   get urlSuffix () {
-    return process.env.ELECTRON_CUSTOM_FILENAME || this.opts.customFilename || this.filename
+    return process.env.ELECTRON_CUSTOM_FILENAME ||
+      process.env.npm_config_electron_custom_filename ||
+      process.env.NPM_CONFIG_ELECTRON_CUSTOM_FILENAME ||
+      process.env.npm_package_config_electron_custom_filename ||
```
benlancaster: Thinking about it, I can see why this may not be all that useful to set in a consuming project's package.json since the filename part is (the only?) platform specific bit.

benlancaster:
Can I get some feedback on this pull request please?

njbmartin:
@MarshallOfSound hate to tag, but it would be great to see these PRs make it in.

njbmartin:
@develar can you take a look at this?

lib/index.js
```diff
@@ -24,19 +24,30 @@ class ElectronDownloader {
   }
 
   get baseUrl () {
-    return process.env.NPM_CONFIG_ELECTRON_MIRROR ||
```
MarshallOfSound: This change makes this a major version bump, if we don't reverse the order of these two environment variables it would only be a minor bump.  Can we try keep this change non-breaking?

njbmartin: I'd imagine this applies to `middleUrl` and `urlSuffix` too?

PR #64, opened 2017-11-28
Fix readme.md: Windows cache location

schwaerz:
A tiny PR:
Today I noticed that the description of the cache location is wrong - at least for windows. Didn't check MacOS or Linux, though.


schwaerz:
My fault...

PR #68, opened 2018-02-27
Add option to disable checksum checking

### LINKED ISSUE #51: disable checksum
Could it be possible to disable checksum when we tell electron-download to download electron from a private location? I'm creating a PR to detail this change

MarshallOfSound:
Fixes #51

Not convinced on adding this but I can see it being useful for at least local Electron dev so 🤔

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -160,6 +160,9 @@ class ElectronDownloader {
   }
 
   get verifyChecksumNeeded () {
+    if (this.opts.disableChecksumSafetyCheck) {
+      return false
+    }
     return semver.gte(this.version, '1.3.2')
   }
 
--- a/lib/index.js
+++ b/lib/index.js
@@ -160,10 +160,7 @@ class ElectronDownloader {
   }
 
   get verifyChecksumNeeded () {
-    if (this.opts.disableChecksumSafetyCheck) {
-      return false
-    }
-    return semver.gte(this.version, '1.3.2')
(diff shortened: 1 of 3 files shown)


readme.md
```diff
@@ -90,6 +90,12 @@ The order of precedence is:
 1. the options given to `download`
 1. defaults
 
+You can also disable checksum validation if you really want to (this is in
+general a bad idea).  Do this by setting `disableChecksumSafetyCheck` to `true`
+in the options object.  Use this only when testing local build of Electron,
+if you have internal builds of Electron you should generate the SHASUMS file
+yourself and let `electron-download` still perform it's hash validations.
```
malept: its

lib/index.js
```diff
@@ -160,6 +160,9 @@ class ElectronDownloader {
   }
 
   get verifyChecksumNeeded () {
+    if (this.opts.disableChecksumSafetyCheck) {
+      return false
+    }
```
malept: If you're going to have a separate if-statement for this, I'd like to see a `console.warn` statement :smile:

MarshallOfSound: I initially had one, then realized no one would ever really see it and all that would happen is we'd get an issue asking to disable the console warning 😆

malept: IMO either we add a `debug()` statement to that effect, or we change the method to:

```javascript
return !this.opts.disableChecksumSafetyCheck && semver.gte(this.version, '1.3.2')
```

PR #69, opened 2018-02-27
Add autoDownload option

### LINKED ISSUE #57: Throw error when cached Electron isn't found instead of downloading by default
I use a private build of Electron that I store on my local machine, and I use `electron-packager` to package my application. To use my private Electron, I set the `cache` option.

Debugging my build scripts was very cumbersome, because `electron-download` gave no indication of whether it could find the Electron build in the cache location or not. If it can't find Electron, it defaults to downloading public releases.

This is not what I want! I don't want the publicly available bits. I'd rather `electron-download` throw an error that I can catch and act on quickly.

electron-userland/electron-packager#187

MarshallOfSound:
Fixes #57

This allows you to disable downloading missing files

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -189,7 +189,7 @@ class ElectronDownloader {
       } else if (this.opts.autoDownload) {
         this.ensureCacheDir(cb)
       } else {
-        cb(new Error('File: "' + this.cachedZip + '" does not exist locally and autoDownload is false'))
+        cb(new Error(`File: "${this.cachedZip}" does not exist locally and autoDownload is false`))
       }
     })
   }
--- a/lib/index.js
+++ b/lib/index.js
@@ -13,7 +13,10 @@ const sumchecker = require('sumchecker')
 
 class ElectronDownloader {
   constructor (opts) {
-    this.opts = opts
+    this.opts = Object.assign({ autoDownload: true }, opts)
+    if (this.opts.force && !this.opts.autoDownload) {
+      throw new Error('force and autoDownload options are incompatible for Electron Download')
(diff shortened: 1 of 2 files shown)


lib/index.js
```diff
@@ -183,8 +186,10 @@ class ElectronDownloader {
       if (exists && !this.force) {
         debug('zip exists', this.cachedZip)
         this.checkIfZipNeedsVerifying(cb)
-      } else {
+      } else if (this.opts.autoDownload) {
         this.ensureCacheDir(cb)
+      } else {
+        cb(new Error('File: "' + this.cachedZip + '" does not exist locally and autoDownload is false'))
```
malept: Can use interpolation syntax here

lib/index.js
```diff
@@ -307,6 +312,10 @@ class ElectronDownloader {
...
 
 module.exports = function download (opts, cb) {
-  const downloader = new ElectronDownloader(opts)
-  downloader.downloadIfNotCached(cb)
+  try {
+    const downloader = new ElectronDownloader(opts)
+    downloader.downloadIfNotCached(cb)
+  } catch (err) {
+    cb(err)
+  }
```
malept: Oof. Would be nice if this module was ported to a Promise-based API.

MarshallOfSound: Yes, yes it would 😆 #70

lib/index.js
MarshallOfSound: Does node 4 support that?

malept: Yup! Electron Packager uses it everywhere.

PR #74, opened 2018-03-20
Add option to download PDB archive

jan-auer:
Similar to the `dsym` option, this adds an option to download PDBs. This is the most straight-forward implementation, though, It might make sense to unify these flags to something like:

`type: 'binary' (default) | 'symbols' | 'dsym' | 'pdb'`


MarshallOfSound:
Closing as `next` that was just merged to master now supports this 👍

PR #75, opened 2018-04-04
Auto switch mirror for Great Firewall (of China)

gucong3000:


fritx:
`async function` was used at https://github.com/gucong3000/in-gfw/blob/dc6796f2f9814414a6a955fa1a7734a5c5d29c26/lib/os.js#L4

which would require `node >= 7.6`
https://www.infoq.com/news/2017/02/node-76-async-await

and break compatibility defined at https://github.com/electron-userland/electron-download/blob/777ca9ff99906ba114e9a1d66717b3d904766cd1/package.json#L60-L62

```json
"engines": {
  "node": ">= 4.0"
}
```

gucong3000:
Node V4 working properly:
https://travis-ci.org/electron-userland/electron-download/jobs/361947356

gucong3000:
> async function was used

No, it is just a function named `async`

MarshallOfSound:
This IMO is not something we want, we already have a well documented and well-used proxy / mirror logic that allows users to set this up themselves.  I don't want to point people somewhere else.

If you feel documentation around the mirror can be improved feel free but it is already referenced in the README of `electron/electron` so it can't get much more prominent.

MarshallOfSound:
Woops, leaving open for discussion

gucong3000:
> we already have a well documented and well-used

Most of the users that affected by Great Firewall speak Chinese. They can't read English.

fritx:
> No, it is just a function named async

Haha... found myself stupid.. and `in-gfw` looks quite awesome.

PR #79, opened 2018-12-04
 Convert to use Promises

### LINKED ISSUE #70: Switch to promise based API
Will be a major bump, maybe we can make this backwards compatible (I.e. support `cb` syntax as well as a promise based syntax)

malept:
This is mostly code reorganization. It adds the `nodeify` module, which should be removed in the next major version bump, and `pify`, to promisify `nugget`. It also removes `path-exists` since the functionality is provided by `fs-extra`.

Fixes #70.


MarshallOfSound:
Closing as `next` is promise based

PR #80, opened 2019-01-22
fix: process raw-passed arm architecture correctly

visleaf:
This patch should improve ARM architecture processing.

For example, if we are using this module to download Electron binaries natively on some machine with an ARM architecture and carelessly providing value of the `process.arch` constant as the `arch` option's value, `electron-download` will try to download Electron binaries for an `arm` architecture. With this patch `electron-download` will try to assemble string containing valid version of an ARM architecture on this machine.


visleaf:
Bringing @malept in. He knows something about ARM, so we can discuss these changes, I hope :)

visleaf:
> I don't understand why you're deleting tests and moving code.

Since refactored code take only about 5 lines I don't see any sense in separating it.

> If anything, you should be adding tests to give examples of how your code changes are supposed to work, because your PR summary isn't clear to me.

I will add tests, thanks!

> As it stands, your code changes will break Electron Packager and Electron Forge, and also cause a major version bump.

Could you please explain why should it break `electron-packager` and `electron-forge`?

visleaf:
@malept, I don't think that it is right to use this feature from other modules. It is like using private fields or methods. Isn't it?

MarshallOfSound:
@visleaf We control both modules, so we know what is considered API in each.  In this case that lib/arch file is considered API

visleaf:
@malept, @MarshallOfSound, I will look for alternative design for this change, which will not be a breaking API and make another pull request. Thank you for quick responses.

PR #81, opened 2019-01-23
fix: process raw-passed arm architecture correctly

visleaf:
In the previous design of this change I took on too much, excuse me :)

This patch slightly improves ARM architecture processing. If you pass the `arm` string as the `arch` option's value (it can happen if you used, for example, `process.arch`), then `electron-download` will determine ARM version of your machine itself using it's `lib/arch.js`.

Bringing @malept and @MarshallOfSound in.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -63,7 +63,7 @@ class ElectronDownloader {
   }
 
   get arch () {
-    return this.opts.arch || arch.host(this.quiet)
+    return (this.opts.arch && this.opts.arch !== 'arm') ? this.opts.arch : arch.host(this.quiet)
   }
 
   get cache () {


malept:
If I'm understanding you correctly, you want to be able to pass in `arm` as the arch and not have `electron-download` try to make it more specific.

I'm hesitant to make this change because I don't want to encourage people to either:
1. download the `arm` versions of the official Electron prebuilt binaries, because `armv7l` is the more correct version and Electron no longer generates binaries with that inaccurate arch name, or
2. tag a custom Electron binary with the inaccurate `arm` arch.
2.

malept:
I don't think I understand. Can you paste some example code with results that occur with and without the pull request applied?

visleaf:
@malept, of course.

Imagine that we are running this code snippets on an ARM machine.
```
> process.platform
'linux'
> process.arch
'arm'
> process.config.variables.arm_version
'7'
```

#### Without this pull request applied
We are trying to download `ffmpeg` **without** the `arch` option specified.
```
'use strict';

const ed = require('electron-download');

ed({
  ffmpeg: true,
  platform: process.platform,
  version: '4.0.0'
}, (error, path) => { console.log(`Error:\n  ${error}\nGot:\n  ${path}`); });
```
```
Error:
  null
Got:
  /home/visleaf/.cache/electron/ffmpeg-v4.0.0-linux-armv7l.zip
```
- - -
We are trying to download `ffmpeg` **with** the `arch` option specified.
```
'use strict';

const ed = require('electron-download');

ed({
  ffmpeg: true,
  platform: process.platform,
  arch: process.arch,
  version: '4.0.0'
}, (error, path) => { console.log(`Error:\n  ${error}\nGot:\n  ${path}`); });
```
```
Error:
  Error: Failed to find Electron v4.0.0 for linux-arm at https://github.com/electron/electron/releases/download/v4.0.0/ffmpeg-v4.0.0-linux-arm.zip
Got:
  undefined

```
#### With this pull request applied
We are trying to download `ffmpeg` **without** the `arch` option specified.
```
'use strict';

const ed = require('electron-download');

ed({
  ffmpeg: true,
  platform: process.platform,
  version: '4.0.0'
}, (error, path) => { console.log(`Error:\n  ${error}\nGot:\n  ${path}`); });
```
```
Error:
  null
Got:
  /home/visleaf/.cache/electron/ffmpeg-v4.0.0-linux-armv7l.zip
```
- - -
We are trying to download `ffmpeg` **with** the `arch` option specified.
```
'use strict';

const ed = require('electron-download');

ed({
  ffmpeg: true,
  platform: process.platform,
  arch: process.arch,
  version: '4.0.0'
}, (error, path) => { console.log(`Error:\n  ${error}\nGot:\n  ${path}`); });
```
```
Error:
  null
Got:
  /home/visleaf/.cache/electron/ffmpeg-v4.0.0-linux-armv7l.zip
```

visleaf:
@malept, excuse me for a long response. What scenario you would like to see?

malept:
Basically, your examples but only testing the `get arch` method. We'd have to start exporting `ElectronDownloader`.

visleaf:
> We'd have to start exporting `ElectronDownloader`.

There is no way to access `arch` at this moment or I'm misunderstood you?

malept:
Currently, there is no way to access `ElectronDownloader` except through the exported `download` function, which runs the entire workflow. I'd like to avoid that.

visleaf:
@malept, how can I test `arch` getter in this case?

malept:
Something along the lines of asserting the values of `new ElectronDownloader(opts).arch` with the appropriate things stubbed. See https://github.com/electron-userland/electron-download/blob/master/test/test_arch.js for tests you can find inspiration from.

visleaf:
I am totally confused by this:
> Currently, there is no way to access `ElectronDownloader` except through the exported `download` function, which runs the entire workflow. I'd like to avoid that.

> Something along the lines of asserting the values of `new ElectronDownloader(opts).arch` with the appropriate things stubbed.

I mean, I don't understand how I can test the `arch` getter if there is no way to instantiate `ElectronDownloader` by myself.

malept:
Add `ElectronDownloader` to the exports:

```javascript
module.exports.ElectronDownloader = ElectronDownloader
```

visleaf:
Closing this since I have absolutely no free time recently. You have made a lot of changes and I'm unable to adapt this pull request to the new codebase now. If you consider these changes important, you may implement them yourself.

PR #88, opened 2019-04-30
chore: upgrade sumchecker to ^3.0.0

malept:
`sumchecker` v3 comes with its own typings file.


malept:
It fails with `tsc -p tsconfig.esm.json` ... :confused: 

Can it not find the module because sumchecker doesn't have an ES module?

PR #89, opened 2019-05-23
fix: add user-friendly validation when an invalid input object is provided

MarshallOfSound:
Those using typescript will get compile errors, we should still have some handy warnings for those not using typescript so they get more descriptive errors than "can not read property startsWith of undefined".


MarshallOfSound:
> Looks fine. Perhaps add a test?

If I add a test I'll have to litter the test file with `any` casts because typescript won't let you not provide these properties 😆

PR #90, opened 2019-05-26
feat: pass in downloader options to got

malept:
This is needed for a lot of functionality that people expect from Electron Packager.

### DIFF
--- a/src/GotDownloader.ts
+++ b/src/GotDownloader.ts
@@ -1,12 +1,11 @@
 import * as fs from 'fs-extra';
 import * as got from 'got';
-import { RequestOptions } from 'https';
 import * as path from 'path';
 
 import { Downloader } from './Downloader';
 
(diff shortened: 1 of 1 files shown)


src/GotDownloader.ts
```diff
@@ -4,12 +4,12 @@ import * as path from 'path';
 
 import { Downloader } from './Downloader';
 
-export class GotDownloader implements Downloader<null> {
-  async download(url: string, targetFilePath: string) {
+export class GotDownloader implements Downloader<any> {
+  async download(url: string, targetFilePath: string, options?: any) {
```
malept: You're probably asking, "Why is this `any`?" I'm using the same signature as `got` itself:

https://github.com/sindresorhus/got/blob/f77db9929096d0b56cde7981e3f8b1ca635eb628/source/create.ts#L60

PR #101, opened 2019-06-19
chore: fix workflow name and restrict to master

malept:
GitHub Actions is very particular about this.


.github/main.workflow
```diff
@@ -3,8 +3,15 @@ workflow "Build & deploy docs" {
   resolves = ["Publish Documentation"]
 }
 
+action "Only Master" {
```
MarshallOfSound: Can this be https://github.com/actions/bin/tree/master/filter instead?

malept: Aha! Yes.

PR #103, opened 2019-06-25
fix: Export all types

ffflorian:
This PR will
- export all types so interfaces like `ElectronDownloadRequestOptions` can be imported.

cc @MarshallOfSound @malept


ffflorian:
:tada:

PR #104, opened 2019-06-26
chore: fix up publishing docs to GitHub repo

malept:
I figured out how to locally test the silly thing and "it works on my machine":tm:

It's mostly the change to `publish:docs`.


malept:
We also might want to disable CircleCI tests on branches where there's no `.circleci/config.yml`, if that's possible.

PR #109, opened 2019-07-24
feat: make platform and arch optional in downloadArtifact

### LINKED ISSUE #107: feat: make "platform" and "arch" optional for downloadArtifact calls
It would make a lot of sense to make those options not required and default to the current platform and arch. Especially considering how non-trivial the current arch value is determined.

MarshallOfSound:
Fixes #107

### DIFF
--- a/test/index.spec.ts
+++ b/test/index.spec.ts
@@ -120,6 +120,13 @@ describe('Public API', () => {
       expect(await fs.readFile(dtsPath, 'utf8')).toContain('declare namespace Electron');
     });
 
+    it('should work default platform/arch', async () => {
+      await downloadArtifact({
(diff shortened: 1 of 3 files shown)


test/index.spec.ts
```diff
@@ -120,6 +120,13 @@ describe('Public API', () => {
       expect(await fs.readFile(dtsPath, 'utf8')).toContain('declare namespace Electron');
     });
 
+    it('should work default platform/arch', async () => {
```
malept: "should work" until someone tries to install, for example, Electron 5 on Linux/mips64el :smile:

MarshallOfSound: 😢

Issue #108, opened 2019-07-24
bug: the SHASUMS256.txt download ignores mirrorOptions


alexeykuzmin:
When a mirror is used a "SHASUMS256.txt" files still gets downloaded from https://github.com/electron/electron/releases/download/VERSION/SHASUMS256.txt

https://github.com/electron/get/blob/9acfe87243ba5c98bea4956e376c80bbd0f45a80/src/index.ts#L103-L110

PR #110, opened 2019-07-24
fix: ensure the SHASUM download call provides mirrorOptions

MarshallOfSound:
Fixes https://github.com/electron/get/issues/108


malept:
Post-release this should get tests.

PR #113, opened 2019-07-31
feat: make the cache to respect an artifact url

alexeykuzmin:
The cache right now ignores multiple mirrors used simultaneously on the same machine.
I can lead to some non-obvious build issues.

This changes the cache files structure to the following:
```
/
  httpsgithub.comelectronelectronreleasesdownloadv5.0.9chromedriver-v5.0.9-darwin-x64.zip /
    chromedriver-v5.0.9-darwin-x64.zip
  httpsmyawesomemirror.comelectronv5.0.9chromedriver-v5.0.9-darwin-x64.zip /
    chromedriver-v5.0.9-darwin-x64.zip  // from a mirror
  httpsgithub.comelectronelectronreleasesdownloadv5.0.9electron.d.ts /
    electron.d.ts
  httpsgithub.comelectronelectronreleasesdownloadv6.0.2electron.d.ts /
    electron.d.ts  // a different version
```

### DIFF
--- a/test/index.spec.ts
+++ b/test/index.spec.ts
@@ -116,7 +116,7 @@ describe('Public API', () => {
         artifactName: 'electron.d.ts',
       });
       expect(await fs.pathExists(dtsPath)).toEqual(true);
-      expect(path.basename(dtsPath)).toEqual('v2.0.9-electron.d.ts');
+      expect(path.basename(dtsPath)).toEqual('electron.d.ts');
       expect(await fs.readFile(dtsPath, 'utf8')).toContain('declare namespace Electron');
     });
 
--- a/package.json
+++ b/package.json
@@ -26,6 +26,7 @@
     "env-paths": "^2.2.0",
     "fs-extra": "^8.1.0",
     "got": "^9.6.0",
+    "sanitize-filename": "^1.6.2",
     "sumchecker": "^3.0.0"
   },
   "devDependencies": {
@@ -35,6 +36,7 @@
     "@types/got": "^9.4.4",
     "@types/jest": "^24.0.13",
     "@types/node": "^12.0.2",
+    "@types/sanitize-filename": "^1.1.28",
     "gh-pages": "^2.0.1",
     "husky": "^2.3.0",
     "jest": "^24.8.0",
--- a/src/Cache.ts
+++ b/src/Cache.ts
@@ -2,6 +2,7 @@ import debug from 'debug';
 import envPaths from 'env-paths';
 import * as fs from 'fs-extra';
 import * as path from 'path';
+import * as sanitize from 'sanitize-filename';
 
 const d = debug('@electron/get:cache');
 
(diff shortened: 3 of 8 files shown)


malept:
Is this a breaking change?

MarshallOfSound:
> Is this a breaking change?

Came here to comment this, I've got two suggestions here.

1. If the URL is the default URL (i.e. not a mirror) we shouldn't use a `url_hash` folder, it should just be top level, that way we can avoid a breaking change in the cache path.
2. Can we use something more identifiable than the cache_path like a path safe version of the mirror URL.

E.g. if the mirror url is `https://mysite.com/myprefix/download/electron.zip` I want the cache path to be this

`/http_mysite_com_myprefix_download_electron/electron.zip`

alexeykuzmin:
1. It doesn't look like a breaking change to me.
A cache miss, even if a file has been previously downloaded and saved in in the cache, is a normal thing. The file could have been removed by OS or manually by a user, and it's ok to redownload it.

2. I wanted to have a reliable way to make sure there won't be any false positive cache hits.
Are you aware of any standard ways to convert a URL to a safe file path? The one that would respect various schemes, basic credentials, and maybe even a search query parameters.

malept:
> A cache miss, even if a file has been previously downloaded and saved in in the cache, is a normal thing. The file could have been removed by OS or manually by a user, and it's ok to redownload it.

But then you have a bunch of largish dead files just hanging around for a user, if they've been working with Electron for a while.

alexeykuzmin:
> But then you have a bunch of largish dead files just hanging around for a user, if they've been working with Electron for a while.

Files will pile up anyway once you start working with newer releases.

Removing stale files from a cache might be an issue, but I think it's out of scope of this change.

malept:
I would prefer that @MarshallOfSound's suggestion 1 be implemented.

> Are you aware of any standard ways to convert a URL to a safe file path? The one that would respect various schemes, basic credentials, and maybe even a search query parameters.

Electron Packager uses [`sanitize-filename`](https://github.com/parshap/node-sanitize-filename)

alexeykuzmin:
@malept , @MarshallOfSound another review?

test/index.spec.ts
```diff
@@ -116,7 +116,7 @@ describe('Public API', () => {
         artifactName: 'electron.d.ts',
       });
       expect(await fs.pathExists(dtsPath)).toEqual(true);
-      expect(path.basename(dtsPath)).toEqual('v2.0.9-electron.d.ts');
```
MarshallOfSound: This is deliberate, all other artifacts have the version name in them.  If this artifact does not it will be duplicated and overwritten right?

alexeykuzmin: It can be only overwritten if the downloader called with `force: true`.
`electron.d.ts` files for other Electron versions/mirrors will be downloaded to other folders.

The goal of this PR is to make sure every downloaded file is saved to the cache with a unique path. A folder name serves as a unique name, and a file names does not matter for caching.

PR #115, opened 2019-08-24
Ignore IntelliJ

ofek:


malept:
Thanks for the pull request, but I think this should be in your global git ignore file instead.

PR #118, opened 2019-09-06
feat: add initializeProxy helper for downstream packages

malept:
Folks are having trouble with using this module behind corporate proxies (example: https://github.com/electron/electron-packager/issues/1019). This pull request adds a method (code borrowed from yeoman) that will add environment variable support. See the new section of the README for details.

Downstream packages I expect to provide PRs for, if this is merged:
* `electron`
* `electron-packager`
* `@electron-forge/cli`
* `fiddle`, I think?

https://electronjs.org/docs/tutorial/installation#proxies will have to be modified slightly once `electron/electron` is updated (not that it's accurate right now anyway...)

### DIFF
--- a/src/proxy.ts
+++ b/src/proxy.ts
@@ -1,6 +1,6 @@
 import * as debug from 'debug';
 
-const d = debug('@electron/got:proxy');
+const d = debug('@electron/get:proxy');
 
 /**
  * Initializes a third-party proxy module for HTTP(S) requests.
--- a/src/proxy.ts
+++ b/src/proxy.ts
@@ -1,0 +1,23 @@
+import * as debug from 'debug';
+
+const d = debug('@electron/got:proxy');
+
+/**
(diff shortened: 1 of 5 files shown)


src/proxy.ts
```diff
@@ -0,0 +1,23 @@
+import * as debug from 'debug';
+
+const d = debug('@electron/got:proxy');
```
MarshallOfSound: `got`?

malept: I've been looking too much at `got` to figure out the proxy stuff :dizzy_face:

malept:
> Is it worth auto calling `initializeProxy` based on an env var to help out dependant CLI tools.

Seems reasonable. Not sure how we'd handle this in Fiddle, though.

MarshallOfSound:
@malept Fiddle being gui / non-cli would just expose something in settings for this 👍

jasunbourne:
@MarshallOfSound Can this be approved?

PR #129, opened 2019-11-25
chore(deps): upgrade sumchecker to 3.0.1

malept:
This fixes the TypeScript definitions so we don't have to use `require()` for `sumchecker`.


MarshallOfSound:
This is now conflicting

malept:
Rebased

PR #138, opened 2019-12-19
test: add support for running only tests which don't use the network

malept:
Proposed replacement for #135 (based on https://github.com/electron/get/pull/135#issuecomment-565788557).

CC: @dsanders11

Closes #135.


src/FixtureDownloader.ts
```diff
@@ -0,0 +1,16 @@
...
+
+const FIXTURE_DIR = path.resolve(__dirname, '../test/fixtures');
+
+export class FixtureDownloader implements Downloader<any> {
+  async download(url: string, targetFilePath: string) {
+    await fs.ensureDir(path.dirname(targetFilePath));
+    const fixtureFile = path.join(FIXTURE_DIR, path.basename(targetFilePath));
+    if (await fs.pathExists(fixtureFile)) {
+      await fs.copy(fixtureFile, targetFilePath);
+    }
```
dsanders11: Should there be an `else` clause which throws an error? Seems like a potential divergence from `GotDownloader`, where one will throw an error on a bad URL and the other won't.

malept: It could, but none of the tests use it currently, and it probably shouldn't. The lack of a fixture isn't something that really needs to be tested, because at that point you're testing the scaffolding.

The `GotDownloader` spec tests for nonexistent URLs, which seems like a better way to do it.

dsanders11: > It could, but none of the tests use it currently, and it probably shouldn't.

I disagree. It's pretty foreseeable that someone may want to write a test that verifies how `downloadArtifact` handles a bad URL. I was tempted to add one while trying to track down an uncaught error in promise from the Electron npm install script, to check that the error wasn't from within `downloadArtifact` itself.

If `FixtureDownloader` doesn't throw in that case, it will lead to unexplained behavior if someone does try to write that test, as they'll get an unexpected result, and will have to spend time tracking it down. Since the swap to `FixtureDownloader` is in `src/` rather than `test/`, it's not immediately clear that a different downloader is being used.

I'd suggest at least a warning be printed in `else`, that way if someone creates a test which would hit a missing fixture, they are at least warned as to what is happening.

> The lack of a fixture isn't something that really needs to be tested, because at that point you're testing the scaffolding.

What would be being tested is how `downloadArtifact` handles a bad URL, not the lack of a fixture.

src/downloader-resolver.ts
```diff
@@ -1,10 +1,15 @@
 import { Downloader } from './Downloader';
 
 export async function getDownloaderForSystem(): Promise<Downloader<any>> {
-  // TODO: Resolve the downloader or default to GotDownloader
-  // Current thoughts are a dot-file traversal for something like
-  // ".electron.downloader" which would be a text file with the name of the
-  // npm module to import() and use as the downloader
-  const { GotDownloader } = await import('./GotDownloader');
-  return new GotDownloader();
+  if (process.env.ELECTRON_GET_TESTSUITE) {
```
dsanders11: What's the benefit to this approach rather than mocking `getDownloaderForSystem` in the tests? I'm personally not a fan of injecting test code into the main code, it seems cleaner if `FixtureDownloader` lived in `test/`. With this approach it's not easily discerned from the test code that there's a `FixtureDownloader`.

malept: It's mostly because I didn't want to add `proxyquire` to the `devDependencies`.

malept: Turns out `proxyquire` is a gigantic pain to deal with when you're using TypeScript and trying to require a TypeScript file. I ended up just setting `downloader` all throughout the `index` spec, which is suboptimal.

PR #147, opened 2020-02-11
 fix: default falsy platform/arch to host values 

malept:
Alternate proposal to #137 
Closes #137

~~Blocked by #141~~

CC: @dsanders11


dsanders11:
@malept, LGTM.

PR #148, opened 2020-02-21
feat: add progress bar & progress callback option

### LINKED ISSUE #124: electron-download issues with cache and other things
I'm having significant issues with electron's postinstall script. For some reason, I am getting extremely low (~1 kbps) from GitHub Releases. I don't know if it's on my end or GitHub's, everything else is working fine for me.

1. According to https://github.com/electron/electron-quick-start/issues/127 , doing `npm install --verbose electron@7.0.0` should log the download progress, but no such thing is happening.

2. After figuring out why the script was hanging, https://medium.com/@muellermarkus/download-electron-releases-from-a-faster-mirror-than-github-741bac15c3c9 pointed me to https://www.npmjs.com/package/electron-download

3. electron-download's homepage redirects to here, but I'm not seeing any reference to a CLI.

4. electron-download works, but it caches the files without the sanitized url in the file path, while @electron/get *does*, which means, it's not finding the files I downloaded manually.

If electron-download is deprecated, the NPM package should be marked as such.

The download should log progress by default, that's how https://github.com/GoogleChrome/puppeteer does it for Chromium; hanging on `node install.js` for 30min with 0 feedback is terrible UX.

malept:
Console progress bar example:

```
> const { download } = require('@electron/get')
> download('8.0.0')
Downloading electron-v8.0.0-linux-x64.zip: [====================================================================================================] 100% ETA: 0.0 seconds
Downloading SHASUMS256.txt: [====================================================================================================] 100% ETA: 0.0 seconds
```

The progress callback `getProgressCallback`'s main use case (along with the `quiet` option) is for use by Electron Fiddle to power its own React-based progress bar.

Fixes #124.

Addresses https://github.com/electron/fiddle/issues/149.


src/GotDownloader.ts
```diff
@@ -1,20 +1,46 @@
...
 export class GotDownloader implements Downloader<any> {
   /**
    * @param options - see [`got#options`](https://github.com/sindresorhus/got#options) for possible keys/values.
+   * Additional options:
+   * * `quiet`: if `true`, disables the console progress bar (the `ELECTRON_GET_NO_PROGRESS`
+   *   environment variable also does this)
+   * * `getProgressCallback`: if defined, triggers every time `got`'s `downloadProgress` event
+   *   callback is triggered.
    */
   async download(url: string, targetFilePath: string, options?: any) {
```
HashimotoYT: nit: Can we add types for options interface?

malept: Done in https://github.com/electron/get/pull/148/commits/eb21c794fff05cff9251a26c215da710446572a5

malept:
@MarshallOfSound I decided to go with
> Make the progress bar only start drawing after 30 seconds

Seems to work OK.

andyzimmerer:
Yea

PR #155, opened 2020-04-07
feat: add environment variable to override version downloaded

### LINKED ISSUE #126: Environment variable usage is not documented
While skimming the source code, I noticed that the mirror URL can also be constructed from environment variables, and not just an options object. What's more, the environment variables take preference over the options object. Not sure if it was intended to leave this little detail out.

Unless that's intended, documentation for this package isn't that great. I'd love to see this improved.

jkleinsc:
This PR adds support for an environment variable, `ELECTRON_CUSTOM_VERSION`, that allows overriding the version downloaded.  I also added some brief documentation on the mirror environment variables that resolves #126.

One use case for this feature is to allow overriding the version of Electron specified in package.json so that CI on Electron apps can test against newer Electron versions.

### DIFF
--- a/test/index.spec.ts
+++ b/test/index.spec.ts
@@ -112,6 +112,19 @@ describe('Public API', () => {
         downloadOptions: downloadOpts,
       });
     });
+
+    it('should download a custom version of a zip file', async () => {
+      process.env.ELECTRON_CUSTOM_VERSION = '2.0.10';
+      const zipPath = await download('2.0.3', {
+        cacheRoot,
+        downloader,
+      });
+      expect(typeof zipPath).toEqual('string');
+      expect(await fs.pathExists(zipPath)).toEqual(true);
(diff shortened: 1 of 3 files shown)


test/index.spec.ts
```diff
@@ -179,5 +192,22 @@ describe('Public API', () => {
...
+        artifactName: 'electron',
+        cacheRoot,
+        downloader,
+        platform: process.env.npm_config_platform || process.platform,
+        arch: process.env.npm_config_arch || process.arch,
+        version: '2.0.3',
+      });
+      expect(typeof zipPath).toEqual('string');
+      expect(await fs.pathExists(zipPath)).toEqual(true);
+      expect(path.basename(zipPath)).toMatchInlineSnapshot(`"electron-v2.0.10-darwin-x64.zip"`);
```
codebytere, line 208: This should invoke `process.platform`, since it runs on all platforms - that's why tests are failing at present

PR #156, opened 2020-05-06
feat: allow optionally only passing a base url to downloadArtifact

codebytere:
In some cases the url passed with a mirror will be fully-formed and not have `path/name` at the end, meaning that the current logic will tack it on and cause failures. This fixes that by optionally allowing only the base url to be passed to got for download.

cc @MarshallOfSound

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,6 +35,16 @@ const zipFilePath = await downloadArtifact({
 
 ### Specifying a mirror
 
+To specify another location to download Electron assets from, the following options are
+available:
+
+* `mirrorOptions` Object
+  * `mirror` String (optional) - The base URL of the mirror to download from.
+  * `nightly_mirror` String (optional) - The Electron nightly-specific mirror URL.
+  * `customDir` String (optional) - The name of the directory to download from, often scoped by version number.
+  * `customFilename` String (optional) - The name of the asset to download.
+  * `baseOnly` Boolean (optional) - Whether or not to download from the base URL only.
+
 Anatomy of a download URL, in terms of `mirrorOptions`:
 
 ```
--- a/README.md
+++ b/README.md
@@ -43,7 +43,7 @@ available:
   * `nightly_mirror` String (optional) - The Electron nightly-specific mirror URL.
   * `customDir` String (optional) - The name of the directory to download from, often scoped by version number.
   * `customFilename` String (optional) - The name of the asset to download.
-  * `baseOnly` Boolean (optional) - Whether or not to download from the base URL only.
+  * `baseOnly` Boolean (optional) - Whether to download from the base URL only.
 
 Anatomy of a download URL, in terms of `mirrorOptions`:
 
--- a/src/artifact-utils.ts
+++ b/src/artifact-utils.ts
@@ -24,7 +24,11 @@ export function getArtifactFileName(details: ElectronArtifactDetails) {
   ].join('-')}.zip`;
 }
 
-function mirrorVar(name: keyof MirrorOptions, options: MirrorOptions, defaultValue: string) {
+function mirrorVar(
+  name: keyof Omit<MirrorOptions, 'baseOnly'>,
+  options: MirrorOptions,
+  defaultValue: string,
+) {
   // Convert camelCase to camel_case for env var reading
   const lowerName = name.replace(/([a-z])([A-Z])/g, (_, a, b) => `${a}_${b}`).toLowerCase();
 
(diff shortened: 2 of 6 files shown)


codebytere:
Looks like it's using TS `v3.2.4` - not sure why since i updated the lockfile.

src/types.ts
```diff
@@ -8,6 +8,7 @@ export interface MirrorOptions {
   mirror?: string;
   customDir?: string;
   customFilename?: string;
+  baseOnly?: boolean;
```
malept, line 11: Can you add some docs for this property? I know that the others don't have docs, but that's an oversight. 😢

codebytere: @malept i think that should do it, let me know if you'd like more!

src/types.ts
```diff
@@ -5,9 +5,26 @@ export interface MirrorOptions {
...
+   * often scoped by version number e.g 'v4.0.4'
+   */
   customDir?: string;
+  /**
+   * The name of the asset to download,
+   * e.g 'electron-v4.0.4-linux-x64.zip'
+   */
   customFilename?: string;
+  /**
+   * Whether or not to download from the base URL only,
```
malept, line 24: Grammatical nit:

```suggestion
   * Whether to download from the base URL only,
```

README.md
```diff
@@ -35,6 +35,16 @@ const zipFilePath = await downloadArtifact({
...
 
+To specify another location to download Electron assets from, the following options are
+available:
+
+* `mirrorOptions` Object
+  * `mirror` String (optional) - The base URL of the mirror to download from.
+  * `nightly_mirror` String (optional) - The Electron nightly-specific mirror URL.
+  * `customDir` String (optional) - The name of the directory to download from, often scoped by version number.
+  * `customFilename` String (optional) - The name of the asset to download.
+  * `baseOnly` Boolean (optional) - Whether or not to download from the base URL only.
```
malept, line 46: Grammatical nit:
```suggestion
  * `baseOnly` Boolean (optional) - Whether to download from the base URL only.
```

test/artifact-utils.spec.ts
```diff
@@ -66,6 +66,21 @@ describe('artifact-utils', () => {
...
 
+    it('should allow for setting only a base url when mirrorOptions.baseOnly is set', () => {
+      expect(
+        getArtifactRemoteURL({
+          arch: 'x64',
+          artifactName: 'electron',
+          mirrorOptions: {
+            mirror: 'https://mirror.example.com',
+            baseOnly: true,
+          },
```
malept, line 77: Add customDir` and `customFilename` here to show that they don't factor into the URL in this case?

codebytere: @malept it's a little weird to me that `nightly_mirror` isn't camelCase - is that intentional or would you be open to my changing that in a follow-up PR?

malept: I don't think it was intentional although you'd have to doublecheck with @MarshallOfSound 😁 

I'm OK with a PR to deprecate with a camel cased version.

README.md
```diff
@@ -35,6 +35,16 @@ const zipFilePath = await downloadArtifact({
...
 
+To specify another location to download Electron assets from, the following options are
+available:
+
+* `mirrorOptions` Object
+  * `mirror` String (optional) - The base URL of the mirror to download from.
+  * `nightly_mirror` String (optional) - The Electron nightly-specific mirror URL.
+  * `customDir` String (optional) - The name of the directory to download from, often scoped by version number.
+  * `customFilename` String (optional) - The name of the asset to download.
+  * `baseOnly` Boolean (optional) - Whether to download from the base URL only.
```
MarshallOfSound, line 46: `baseOnly` as an API doesn't make a lot of sense when their aren't any other options called "base".  My suggestion here would be to make it take a `resolveAssetURL` method instead which can then return a full URL if it wants.  This gives folks absolute control over the URL that is returned and removed the strangeness around "what is base" and such

src/artifact-utils.ts
```diff
@@ -50,5 +54,5 @@ export function getArtifactRemoteURL(details: ElectronArtifactDetails): string {
   );
   const file = mirrorVar('customFilename', opts, getArtifactFileName(details));
 
-  return `${base}${path}/${file}`;
+  return opts.baseOnly ? `${base}` : `${base}${path}/${file}`;
```
MarshallOfSound, line 57: `base` is already a string here

PR #157, opened 2020-05-06
fix: deprecate mirrorOptions.nightly_mirror in favor of mirrorOptions.nightlyMirror

codebytere:
Fixes a variable inconsistency in `mirrorOptions` - all variables were camelCase except `nightly_mirror`.

cc @malept

### DIFF
--- a/src/artifact-utils.ts
+++ b/src/artifact-utils.ts
@@ -46,7 +46,12 @@ export async function getArtifactRemoteURL(details: ElectronArtifactDetails): Pr
   const opts: MirrorOptions = details.mirrorOptions || {};
   let base = mirrorVar('mirror', opts, BASE_URL);
   if (details.version.includes('nightly')) {
-    base = mirrorVar('nightly_mirror', opts, NIGHTLY_BASE_URL);
+    if (opts.nightly_mirror) {
+      base = mirrorVar('nightly_mirror', opts, NIGHTLY_BASE_URL);
+      console.warn(`nightly_mirror is deprecated, please use nightlyMirror`);
+    } else {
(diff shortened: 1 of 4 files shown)


src/artifact-utils.ts
```diff
@@ -46,7 +46,12 @@ export function getArtifactRemoteURL(details: ElectronArtifactDetails): string {
   const opts: MirrorOptions = details.mirrorOptions || {};
   let base = mirrorVar('mirror', opts, BASE_URL);
   if (details.version.includes('nightly')) {
-    base = mirrorVar('nightly_mirror', opts, NIGHTLY_BASE_URL);
+    if (opts['nightly_mirror']) {
```
malept, line 49:
```suggestion
    if (opts.nightly_mirror) {
```

codebytere:
Should be good now i think!

malept:
Can we keep the `nightly_mirror` test in addition to testing `nightlyMirror`? We can get rid of it once we remove the deprecated option.

src/artifact-utils.ts
```diff
@@ -46,7 +46,12 @@ export async function getArtifactRemoteURL(details: ElectronArtifactDetails): Pr
   const opts: MirrorOptions = details.mirrorOptions || {};
   let base = mirrorVar('mirror', opts, BASE_URL);
   if (details.version.includes('nightly')) {
-    base = mirrorVar('nightly_mirror', opts, NIGHTLY_BASE_URL);
+    if (opts.nightly_mirror) {
```
MarshallOfSound, line 49: This won't determine if they're using the old env var.  You should check if mirrorVar is truthy rather than just checking this option

PR #158, opened 2020-05-06
feat: allow custom URL resolution with resolveAssetURL

codebytere:
Follow-up to https://github.com/electron/get/pull/156.

Refs https://github.com/electron/get/pull/156#discussion_r420974703.

Allow custom URL resolution with `resolveAssetURL`.

cc @MarshallOfSound

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -24,7 +24,7 @@ export interface MirrorOptions {
    * A function allowing customization of the url returned
    * from getArtifactRemoteURL().
    */
-  resolveAssetURL?: (opts: Omit<MirrorOptions, 'resolveAssetURL'>) => Promise<string>;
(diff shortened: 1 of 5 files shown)


src/types.ts
```diff
@@ -21,10 +21,10 @@ export interface MirrorOptions {
    */
   customFilename?: string;
   /**
-   * Whether to download from the base URL only,
-   * ignoring customDir and customFilename
+   * A function allowing customization of the url returned
+   * from getArtifactRemoteURL().
    */
-  baseOnly?: boolean;
+  resolveAssetURL?: (opts: Omit<MirrorOptions, 'resolveAssetURL'>) => Promise<string>;
```
MarshallOfSound, line 27: We should probably give it the full options object tbh.  The URL resolver will need to know platform / version / arch, etc.

PR #182, opened 2021-01-20
fix: skip checksum for versions without shasums256

### LINKED ISSUE #181: Unable to download old releases without SHASUMS256.txt
Versions of Electron prior to v1.3.2 did not come with a `SHASUMS256.txt` artifact, but `@electron/get` will always attempt to calculate the checksum regardless.

Ref https://github.com/electron/get/pull/25/ for the old `electron-download` implementation.

erickzhao:
Fixes #181.

Reimplementation of #25 for `@electron/get`. Adds the `semver` package to check version ranges. Note that semver@6 was used because this module still specifies Node 8.6 support.

🧪 Tests: I don't think existing tests referenced the `sumchecker` module, so I haven't added any tests. Let me know if I should!

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -2,6 +2,7 @@ import * as childProcess from 'child_process';
 import * as fs from 'fs-extra';
 import * as os from 'os';
 import * as path from 'path';
+import * as semver from 'semver';
 
 async function useAndRemoveDirectory<T>(
   directory: string,
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -2,7 +2,6 @@ import * as childProcess from 'child_process';
 import * as fs from 'fs-extra';
 import * as os from 'os';
 import * as path from 'path';
-import * as semver from 'semver';
 
 async function useAndRemoveDirectory<T>(
   directory: string,
--- a/package.json
+++ b/package.json
@@ -30,6 +30,7 @@
     "fs-extra": "^8.1.0",
     "got": "^9.6.0",
     "progress": "^2.0.3",
+    "semver": "^6.2.0",
     "sumchecker": "^3.0.1"
   },
   "devDependencies": {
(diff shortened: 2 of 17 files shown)


malept (changes requested):
Thanks for working on this!

I would add a test to make sure that `downloadArtifact` doesn't get called for the super old Electron versions.

erickzhao:
> I would add a test to make sure that downloadArtifact doesn't get called for the super old Electron versions.

Done! I checked for `sumchecker` because I didn't know how to mock for recursive calls of `downloadArtifact()`.

I didn't do exhaustive tests in the boundaries for the npm versions for the sake of not adding too many fixtures, but let me know if it's preferable to do it that way (i.e. test for <1.3.1, 1.3.1, 1.3.2, 1.3.3, 1.3.4, 1.3.5, >1.3.5).

src/utils.ts
```diff
@@ -2,6 +2,7 @@ import * as childProcess from 'child_process';
 import * as fs from 'fs-extra';
 import * as os from 'os';
 import * as path from 'path';
+import * as semver from 'semver';
```
malept, line 5: Unused import?

_(Side note: is prettier even run in CI? Why isn't eslint in this repo?? This should get fixed in a separate PR :sob:)_

malept (approved):
Exhaustive testing is not needed, just test each branch (no checksum file, badly written checksum file :grimacing:, correct checksum file) which you've already done.

This looks good to me modulo a manual linting warning.

src/utils.ts
erickzhao: Prettier is run on `test` as well, not sure why it didn't catch it!

From Circle:
```
> @electron/get@0.0.0-development lint /Users/distiller/project
> prettier --check "src/**/*.ts"

Checking formatting...
All matched files use Prettier code style!
```

malept: > Prettier is run on `test` as well

:facepalm: I did not scroll in the test output

> not sure why it didn't catch it!

Unused imports isn't a formatting thing, it's a linter thing. Hence my comment about `eslint`.

PR #184, opened 2021-01-21
chore: add eslint

malept:
Follow-up to https://github.com/electron/get/pull/182#discussion_r561418583. It's mostly moving code around, adding/clarifying type annotations, and ignoring a bunch of `any`s :grimacing: 

Requires #183.

### DIFF
--- a/package.json
+++ b/package.json
@@ -96,5 +96,8 @@
   "optionalDependencies": {
     "global-agent": "^2.0.2",
     "global-tunnel-ng": "^2.7.1"
+  },
+  "resolutions": {
+    "*/@typescript-eslint/typescript-estree/semver": "^6.3.0"
   }
 }
--- a/package.json
+++ b/package.json
@@ -48,6 +48,7 @@
     "@typescript-eslint/parser": "^2.34.0",
     "eslint": "^6.8.0",
     "eslint-config-prettier": "^6.15.0",
+    "eslint-plugin-import": "^2.22.1",
     "eslint-plugin-jest": "^24.1.3",
     "husky": "^2.3.0",
     "jest": "^24.8.0",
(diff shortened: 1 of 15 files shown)


package.json
```diff
@@ -51,17 +59,32 @@
...
   },
   "husky": {
     "hooks": {
       "pre-commit": "lint-staged"
     }
   },
+  "lint-staged": {
+    "*.ts": [
+      "eslint --fix",
+      "prettier --write",
```
erickzhao, line 84: Question: Is it necessary to have ESLint and Prettier running given that `eslint-config-prettier` is set up now?

ckerr: I'm not positive but I _think_ explicit invocation of both is still useful as-is in this PR.

If we wanted to combine the steps we could use `eslint-plugin-prettier`

test/utils.spec.ts
```diff
@@ -96,24 +99,27 @@ describe('utils', () => {
...
       });
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
       process.config.variables = {} as any;
       expect(getHostArch()).toEqual('armv7l');
     });
 
-    it('should return uname on arm 6', () => {
-      if (process.platform !== 'win32') {
+    if (process.platform !== 'win32') {
+      it('should return uname on arm 6', () => {
```
ckerr, line 108: :+1: on reversing this order

Just curious, was the linter smart enough to catch this, or did you spot it while adding the eslint-disable-next-line comments?

ckerr (approved):
LGTM pending confirmation on @erickzhao's question

package.json
erickzhao: Ah, makes sense to me, I got `eslint-plugin-prettier` and `eslint-config-prettier` mixed up in my head. Thank you!

test/utils.spec.ts
malept: IIRC the linter (`eslint-plugin-jest`) was smart enough, it didn't like conditional `expect`s.

PR #186, opened 2021-01-25
fix: avoid long paths in cache & lower the number of folders created

### LINKED ISSUE #185: Easily hitting long file paths on Windows
It's extremely easy to hit long file paths errors in the cache on Windows:

https://github.com/electron/get/blob/42d964b9bcc2be1d10b472f20ae6395f7423dd76/src/Cache.ts#L21

We should rather come up with something different here, in order to make the file path shorter: ie take the filename of the URL (for easy cache debugging) as well as a short hash checksum of the entire URL (for uniqueness).

joaomoreno:
Fixes #185 by using a checksum of the stripped URL instead of a filenamify version of it.

### DIFF
--- a/src/Cache.ts
+++ b/src/Cache.ts
@@ -1,9 +1,9 @@
 import debug from 'debug';
 import envPaths from 'env-paths';
-import * as filenamify from 'filenamify';
 import * as fs from 'fs-extra';
 import * as path from 'path';
 import * as url from 'url';
+import * as crypto from 'crypto';
 
 const d = debug('@electron/get:cache');
 
@@ -15,11 +15,15 @@ export class Cache {
   constructor(private cacheRoot = defaultCacheRoot) {}
 
   public getCachePath(downloadUrl: string, fileName: string): string {
-    const { search, hash, ...rest } = url.parse(downloadUrl);
+    const parsedDownloadUrl = url.parse(downloadUrl);
+    const { search, hash, ...rest } = parsedDownloadUrl;
     const strippedUrl = url.format(rest);
+    const checksum = crypto
+      .createHash('sha256')
+      .update(strippedUrl)
+      .digest('hex');
 
-    const sanitizedUrl = filenamify(strippedUrl, { maxLength: 255, replacement: '' });
-    return path.resolve(this.cacheRoot, sanitizedUrl, fileName);
+    return path.resolve(this.cacheRoot, checksum, fileName);
   }
 
   public async getPathForFileInCache(url: string, fileName: string): Promise<string | null> {
--- a/src/Cache.ts
+++ b/src/Cache.ts
@@ -23,8 +23,7 @@ export class Cache {
     return crypto
       .createHash('sha256')
       .update(strippedUrl)
-      .digest('hex')
-      .substr(0, 10);
+      .digest('hex');
   }
 
   public getCachePath(downloadUrl: string, fileName: string): string {
--- a/src/Cache.ts
+++ b/src/Cache.ts
@@ -23,7 +23,8 @@ export class Cache {
     return crypto
       .createHash('sha256')
       .update(strippedUrl)
-      .digest('hex');
+      .digest('hex')
+      .substr(0, 10);
   }
 
   public getCachePath(downloadUrl: string, fileName: string): string {
--- a/src/Cache.ts
+++ b/src/Cache.ts
@@ -14,17 +14,20 @@ const defaultCacheRoot = envPaths('electron', {
 export class Cache {
   constructor(private cacheRoot = defaultCacheRoot) {}
 
-  public getCachePath(downloadUrl: string, fileName: string): string {
+  public static getCacheDirectory(downloadUrl: string): string {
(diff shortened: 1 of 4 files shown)


ckerr (approved):
Looks reasonable to me. :+1: on using a checksum; that should avoid collisions that could occur from long filenames being truncated to some non-unique segment.

@joaomoreno could you push another empty commit with a semantic comment to make the Semantic Pull Request bot happy :slightly_smiling_face:

erickzhao (approved):

joaomoreno:
> @joaomoreno could you push another empty commit with a semantic comment to make the Semantic Pull Request bot happy 🙂

Not knowing what Semantic Pull Requests are, believe me I tried. I read [this](https://github.com/zeke/semantic-pull-requests) and I still couldn't figure out what I could do to make this a semantic PR. There's nothing in that link that describes what exactly makes a commit or a PR semantic. After some more clicking, I finally [got it](https://www.conventionalcommits.org/en/v1.0.0/).

Having read that, I love it. ❤️

joaomoreno:
Merged upstream changes, since there were conflicts. Also hit this: https://github.com/electron/get/issues/187

malept (changes requested):
This is a good idea, thanks for making the PR for it.

That being said, I'm blocking this mostly because since we're technically breaking the cache path semantics anyway, I would like to change it so that it doesn't generate two file system nodes for every cached file. Here's what I mean:

Current semantics:

```
$CACHE_DIR/$ESCAPED_URL/basename($URL)
```

This PR, currently:

```
$CACHE_DIR/sha256($URL)/basename($URL)
```

This basically means that for every artifact downloaded via `@electron/get`, a unique folder and file are created. This can add up and cause file indexing from your OS to get rather unhappy.

My proposal:

```
$CACHE_DIR/sha256(dirname($URL))/basename($URL)
```

This way, `$CACHE_DIR/sha256(dirname($URL))` will contain all of the zip files for a given Electron version (even when platform/arch differ) plus its respective checksum file.

joaomoreno:
@malept Sounds good. Would you like me to push those changes here as well?

malept:
That would be great!

joaomoreno:
@malept Done. I've also reconsidered the length of the SHA256 hashes and think that trimming them to 10 chars is good enough for our purposes.

src/Cache.ts
```diff
@@ -14,13 +14,21 @@ const defaultCacheRoot = envPaths('electron', {
...
-    const { search, hash, ...rest } = url.parse(downloadUrl);
-    const strippedUrl = url.format(rest);
+    const { search, hash, pathname, ...rest } = parsedDownloadUrl;
+    const strippedUrl = url.format({ ...rest, pathname: path.dirname(pathname || 'electron') });
+
+    return crypto
+      .createHash('sha256')
+      .update(strippedUrl)
+      .digest('hex')
+      .substr(0, 10);
```
malept, line 27: I hope we don't hit any collisions with 10 characters. The main file I'm worried about is the checksum file.

test/Cache.spec.ts
```diff
@@ -1,15 +1,14 @@
...
 import { Cache } from '../src/Cache';
 
 describe('Cache', () => {
   let cacheDir: string;
   let cache: Cache;
 
-  const dummyUrl = 'dummy://';
-  const sanitizedDummyUrl = 'dummy';
+  const dummyUrl = 'dummy://dummypath';
+  const sanitizedDummyUrl = Cache.getCacheDirectory(dummyUrl);
```
malept, line 11: I think this value is supposed to be hardcoded, since it's testing the implementation.

src/Cache.ts
joaomoreno: We still get `16^10`, or `1099511627780` different checksums. For two different URLs to match they have to match the same checksum *and* the same file name. I think we'll be fine. 😉

joaomoreno:
Addressed feedback.

src/Cache.ts
malept: 16^10 is the problem space, sure, but the hash algorithm could somehow favor one prefix over another (because overall, the hash is "unique", but prefixes aren't as guaranteed).

The checksum file does have the same name for every version, that's why I called it out.

joaomoreno: I meant `electron-v9.4.2-win32-arm64.zip` as a file name, not the checksum. I'd love to get this in asap to address issues downstream. So if this stripping isn't good enough let me know and I can revert it.

joaomoreno:
Reverted this change. Can we publish this any time soon? We are dependent on it.

joaomoreno:
@malept Can we move this forward?

joaomoreno:
@malept Can we get this released?

joaomoreno:
@malept @ckerr Can we get this out? It's preventing us from clearing some security vulnerability alerts downstream.

malept (approved):
Thanks for your patience, this will go out shortly.

joaomoreno:
@malept No worries, I imagine you must be pretty busy as well. Thanks! 🙏

PR #195, opened 2021-06-01
chore: update dependencies

ahippler:
BREAKING CHANGE: drop node 8 and 10 support

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -21,13 +21,13 @@ version: 2.1
 orbs:
   win: circleci/windows@1.0.0
 jobs:
-  test-linux-8:
+  test-linux-12:
     docker:
-      - image: circleci/node:8
+      - image: circleci/node:12
     <<: *steps-test
-  test-linux-10:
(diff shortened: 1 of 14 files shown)


malept (changes requested):
We appreciate the effort here, but as a policy we do not accept `yarn.lock` changes from untrusted contributors (that is, non-Dependabot, non-members of the Electron org). See https://snyk.io/blog/why-npm-lockfiles-can-be-a-security-blindspot-for-injecting-malicious-modules/ for details.

There also appears to be some non-trivial changes which need some explanation as to why they were made.

ahippler:
@malept how do you want to proceed?

malept:
The Ecosystem WG talked about it and we're going to pass on this PR for the reasons I listed earlier.

gjasny:
And who's going to updates the dependencies when?

malept:
> And who's going to update the dependencies and when?

It's on my list, but it's a long list and I do this in my increasingly limited spare time.

PR #198, opened 2021-08-02
feat: add ability to override cache location

ckerr:
I'm writing tests in another repo that inject fake downloads into the process and don't want them to pollute the real cache on my system. Also, I don't want the real cache's checksums to cause my tests to fail because they don't match the fakes' checksums.

The workflow I have in mind is that the test setup code will create a tmpdir and the set `ELECTRON_GET_CACHE_DIR` to point at that tmpdir.

### DIFF
--- a/README.md
+++ b/README.md
@@ -115,6 +115,7 @@ This module downloads Electron to a known place on your system and caches it
 so that future requests for that asset can be returned instantly.  The cache
 locations are:
 
+* If the environment variable `ELECTRON_GET_CACHE_DIR` is set, that path is used.
 * Linux: `$XDG_CACHE_HOME` or `~/.cache/electron/`
 * MacOS: `~/Library/Caches/electron/`
 * Windows: `%LOCALAPPDATA%/electron/Cache` or `~/AppData/Local/electron/Cache/`
--- a/src/Cache.ts
+++ b/src/Cache.ts
@@ -7,9 +7,11 @@ import * as crypto from 'crypto';
 
 const d = debug('@electron/get:cache');
 
-const defaultCacheRoot = envPaths('electron', {
-  suffix: '',
-}).cache;
+const defaultCacheRoot =
+  process.env.ELECTRON_GET_CACHE_DIR ||
+  envPaths('electron', {
+    suffix: '',
+  }).cache;
 
 export class Cache {
   constructor(private cacheRoot = defaultCacheRoot) {}


src/Cache.ts
```diff
@@ -7,9 +7,11 @@ import * as crypto from 'crypto';
 
 const d = debug('@electron/get:cache');
 
-const defaultCacheRoot = envPaths('electron', {
-  suffix: '',
-}).cache;
+const defaultCacheRoot =
+  process.env.ELECTRON_GET_CACHE_DIR ||
```
malept, line 11: Do we need to be able to specify this via npm config like several other configuration options?

https://github.com/electron/get/blob/519d3cc56ae5e553a4952f5f687f094599cf473a/src/artifact-utils.ts#L35-L42

malept (commented):
I assume you're going to be adding tests once we settle on a design?

README.md
```diff
@@ -115,6 +115,7 @@ This module downloads Electron to a known place on your system and caches it
 so that future requests for that asset can be returned instantly.  The cache
 locations are:
 
+* If the environment variable `ELECTRON_GET_CACHE_DIR` is set, that path is used.
```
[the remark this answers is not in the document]
ckerr: That setting appears to be `electron/electron` only (in [npm/install.js](https://github.com/electron/electron/blame/71697344706f39ad12ad20c21f1c3cb86f8e5228/npm/install.js#L43) and [script/release/publish-to-npm.js](https://github.com/electron/electron/blame/71697344706f39ad12ad20c21f1c3cb86f8e5228/script/release/publish-to-npm.js#L150); I don't see any hits for it in `electron/got`.

Also, I don't recall the last time I saw a lowercase environment variable :slightly_smiling_face: and it seems like converting those docs and those two uses in `e/e` smells less bad than propagating that code smell into `electron/got`. However, would converting it `ELECTRON_GET_CACHE_DIR` technically be a breaking change as `electron_config_cache` is a documented environment variable?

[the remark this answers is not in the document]
malept: I'd say it's not a breaking change because it broke when we went from `electron-download` to `@electron/get` a while ago.

src/Cache.ts
ckerr: I'd be happy for a second opinion on this because I'm not so familiar with this codebase, but -- at first read, I don't think so? It looks to me like line 39 in that snippet will cover any environment variable beginning with `ELECTRON_`, including this one.

malept: Well, we'd have to use that method in this callsite if we want to use functionality I linked. Unless I'm misunderstanding you.

ckerr: It's more likely that I'm the one misunderstanding, as I've not read `artifact-utils.ts` before.

**If** I'm reading this file correctly, it looks like mirrorVar() is used to build the URL path for a remote artifact and so a per-client destination directory setting would not need to be mapped. The server is just serving files; it doesn't need to know or care where the client stores them. There's no reason for its URLs' path to have any mapping to a per-client setting...?

README.md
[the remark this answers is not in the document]
ckerr: OK. Then if / when this merges I'll create a followup in `e/e` to update the use there.

src/Cache.ts
```diff
@@ -7,9 +7,11 @@ import * as crypto from 'crypto';
...
+  envPaths('electron', {
+    suffix: '',
+  }).cache;
```
malept, line 14: Here's my proposal: 
`mirrorVar` should be renamed to something like `userCustomValue`, it's not particularly specific to the mirror code _(except for the `MirrorOptions` part which can be genericized to `Record<string, unknown>`)_, and perhaps move it to a different file. Then we can change this to:
```suggestion
const defaultCacheRoot =
  userCustomValue(
    'cacheDir',
    {}, // this is set later on in the Cache constructor, I think, so passing a Record<string, unknown> isn't needed
    envPaths('electron', {
      suffix: '',
    }).cache
  );
```

shenzhuxi:
https://www.electron.build/multi-platform-build.html The document also needs a update. 
"ELECTRON_CACHE" has been replaced by "electron_config_cache", and it's also case sensitive.

erickzhao:
@shenzhuxi this would need to be raised with the maintainers of https://github.com/electron-userland/electron-builder.

PR #200, opened 2021-08-18
feat: support hardcoded checksums

MarshallOfSound:
We want to add the checksums directly to the published NPM package.  To do so `@electron/get` needs to be able to take the hardcoded list and pass it to sumchecker.  This makes the security chain of our checksums

yarn.lock integrity --> electron npm package --> hardcoded checksums --> remote `electron` binary is validated

### DIFF
--- a/src/types.ts
+++ b/src/types.ts
@@ -64,7 +64,7 @@ export interface ElectronDownloadRequestOptions {
   /**
    * Provides checksums for the artifact as strings.
    * Can be used if you already know the checksums of the Electron artifact
-   * you are downloading and want to skip the checksum file download for
+   * you are downloading and want to skip the checksum file download
    * without skipping the checksum validation.
    *
    * This should be an object whose keys are the file names of the artifacts and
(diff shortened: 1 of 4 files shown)


src/types.ts
```diff
@@ -61,6 +61,16 @@ export interface ElectronDownloadRequestOptions {
    * Defaults to `false`.
    */
   unsafelyDisableChecksums?: boolean;
+  /**
+   * Provides checksums for the artifact as strings.
+   * Can be used if you already know the checksums of the Electron artifact
+   * you are downloading and want to skip the checksum file download for
+   * without skipping the checksum validation.
```
erickzhao, line 68:
```suggestion
   * you are downloading and want to skip the checksum file download
   * without skipping the checksum validation.
```

erickzhao (approved):
Reading through the core logic makes sense to me.

malept:
It looks like this caused a regression: https://github.com/electron/electron-packager/runs/3401225023?check_suite_focus=true

ChALkeR:
This is great, huge thanks for this!
The caching logic has a [bug](https://github.com/electron/get/issues/210) though.

PR #212, opened 2022-01-10
fix: validate checksum of cached artifacts too

### LINKED ISSUE #210: Shasum check is not robust
_First of all, it's great to see that the longstanding issue of trusting a remote SHASUMS256.txt file has been fixed. Huge thanks for that!_

There is a bug in the impl though: it doesn't recheck that shasums match when it pulls a file out of cache.
That happens in https://github.com/electron/get/blob/main/src/index.ts#L72-L82 -- that codepath returns the file from cache without any shasum verification, regardless of how it got into the cache.

### Steps to reproduce, variant 1
1. `npm i electron`
2. replace the cache file (look in `~/.cache/electron`) with something else
3. `npm i electron` in another dir -- the tampered file will be used without any shasum check

### Steps to reproduce, variant 2
1. `npm i electron` to get npm packages
2. `rm -rf ~/.cache/electron` to clear cache
3. Replace the hash in `node_modules/electron/checksums.json` to simulate a hash mismatch
4. `rm -rf ~/node_modules/electron/dist` to trigger reinstall
5. `node ./node_modules/electron/install.js` now fails, as it should
6. `electron_use_remote_checksums=1 node ./node_modules/electron/install.js` succeeds (expected)
7.  `rm -rf ~/node_modules/electron/dist` to trigger reinstall
8. `node ./node_modules/electron/install.js` succeeds, despite the hash mismatching the expected one in `checksums.json`.

I.e. if the first call to whatever put the package in cache has been done e.g. with `electron_use_remote_checksums` env var, or without `checksums` option, or with `unsafelyDisableChecksums` option, the follow up call which has `checksums` option set and expects them to be validated against those supplied ones _won't_ get them validated.

Same as when the cache has been corrupted or tampered, allowing e.g. a long-term CI poisoning via cache artifacts.

MarshallOfSound:
Fixes #210

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -34,6 +34,69 @@ if (process.env.ELECTRON_GET_USE_PROXY) {
   initializeProxy();
 }
 
+type ArtifactDownloader = (
+  _artifactDetails: ElectronPlatformArtifactDetailsWithDefaults,
+) => Promise<string>;
+
+async function validateArtifact(
+  artifactDetails: ElectronArtifactDetails,
+  downloadedAssetPath: string,
+  _downloadArtifact: ArtifactDownloader,
+) {
+  return await withTempDirectoryIn(artifactDetails.tempDirectory, async tempFolder => {
+    // Don't try to verify the hash of the hash file itself
+    // and for older versions that don't have a SHASUMS256.txt
+    if (
+      !artifactDetails.artifactName.startsWith('SHASUMS256') &&
+      !artifactDetails.unsafelyDisableChecksums &&
+      semver.gte(artifactDetails.version, '1.3.2')
+    ) {
+      let shasumPath: string;
+      const checksums = artifactDetails.checksums;
+      if (checksums) {
+        shasumPath = path.resolve(tempFolder, 'SHASUMS256.txt');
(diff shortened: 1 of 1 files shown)


ckerr (approved):
I'm not up-to-speed with all the differences between pre-1.3.2 and post-1.3.4, but the code here looks reasonable

erickzhao (approved):

Toinane (approved):

kanru:
The new behavior seems to be always trying to download the checksum file even if all other files are cached.
I'm unable to build electron apps in a sandboxed builder (without network) due to this.
Could you let me know if I need to cache everything where should I put the `SHASUMS256.txt` file?

MarshallOfSound:
> The new behavior seems to be always trying to download the checksum file even if all other files are cached.

This is the new correct/expected behavior.

> Could you let me know if I need to cache everything where should I put the SHASUMS256.txt file?

Right next to the other files you were pre-filling the cache with

Worth noting that if you want to do this truly offline without prefilling the SHASUM file you could also provide a `checksums` object directly to `@electron/get`.  This is for instance what the core `electron` npm package does --> https://github.com/electron/electron/blob/main/npm/install.js#L45

PR #214, opened 2022-01-25
feat(proxy): support http(s)_proxy sans prefix on nodejs@>=10

BlackHole1:
see: https://github.com/gajus/global-agent/tree/f82ad8a151af30f5bf218487d006771c6407bf89#what-is-the-reason-global-agentbootstrap-does-not-use-http_proxy

### What is the reason `global-agent/bootstrap` does not use `HTTP_PROXY`?

Some libraries (e.g. [`request`](https://npmjs.org/package/request)) change their behaviour when `HTTP_PROXY` environment variable is present. Using a namespaced environment variable prevents conflicting library behaviour.

You can override this behaviour by configuring `GLOBAL_AGENT_ENVIRONMENT_VARIABLE_NAMESPACE` variable, e.g.

```bash
$ export GLOBAL_AGENT_ENVIRONMENT_VARIABLE_NAMESPACE=

```

Now script initialized using `global-agent/bootstrap` will use `HTTP_PROXY`, `HTTPS_PROXY` and `NO_PROXY` environment variables.

----

So `ELECTRON_GET_USE_PROXY` was actually not working until now...

This PR will contribute to the improvement of upstream applications, e.g: [electron-fiddle](https://github.com/electron/fiddle)

close: https://github.com/electron/fiddle/issues/258 https://github.com/electron/fiddle/issues/270

cc: @malept @erickzhao

### DIFF
--- a/src/proxy.ts
+++ b/src/proxy.ts
@@ -12,7 +12,9 @@ export function initializeProxy(): void {
 
     if (MAJOR_NODEJS_VERSION >= 10) {
       // `global-agent` works with Node.js v10 and above.
-      require('global-agent').bootstrap();
+      require('global-agent').bootstrap({
+        environmentVariableNamespace: '',
+      });
     } else {
       // `global-tunnel-ng` works with Node.js v10 and below.
       require('global-tunnel-ng').initialize();
--- a/src/proxy.ts
+++ b/src/proxy.ts
@@ -12,7 +12,7 @@ export function initializeProxy(): void {
     const MAJOR_NODEJS_VERSION = parseInt(process.version.slice(1).split('.')[0], 10);
 
     if (MAJOR_NODEJS_VERSION >= 10) {
-      // See: https://github.com/electron/get/pull/214
+      // See: https://github.com/electron/get/pull/214#discussion_r798845713
       const env = getEnv('GLOBAL_AGENT_');
       process.env.GLOBAL_AGENT_HTTP_PROXY = env('HTTP_PROXY');
       process.env.GLOBAL_AGENT_HTTPS_PROXY = env('HTTPS_PROXY');
--- a/src/proxy.ts
+++ b/src/proxy.ts
@@ -3,6 +3,25 @@ import * as debug from 'debug';
 const d = debug('@electron/get:proxy');
 
 /**
+ * get process.env variable
+ * ignore case and compatible `GLOBAL_AGENT_` namespace
+ */
+function getProxyEnv(name: string): string | undefined {
+  const namespace = 'GLOBAL_AGENT_';
+  const namespaceProxyName = `${namespace}${name}`;
+
+  for (const envName in process.env) {
+    if (namespaceProxyName.toLowerCase() === envName.toLowerCase()) {
+      return process.env[envName];
+    }
+
+    if (name.toLowerCase() === envName.toLowerCase()) {
+      return process.env[envName];
+    }
+  }
+}
+
+/**
  * Initializes a third-party proxy module for HTTP(S) requests.
  */
 export function initializeProxy(): void {
@@ -11,10 +30,13 @@ export function initializeProxy(): void {
     const MAJOR_NODEJS_VERSION = parseInt(process.version.slice(1).split('.')[0], 10);
 
     if (MAJOR_NODEJS_VERSION >= 10) {
+      // See: https://github.com/electron/get/pull/214
+      process.env.GLOBAL_AGENT_HTTP_PROXY = getProxyEnv('HTTP_PROXY');
+      process.env.GLOBAL_AGENT_HTTPS_PROXY = getProxyEnv('HTTPS_PROXY');
+      process.env.GLOBAL_AGENT_NO_PROXY = getProxyEnv('NO_PROXY');
+
       // `global-agent` works with Node.js v10 and above.
-      require('global-agent').bootstrap({
-        environmentVariableNamespace: '',
-      });
+      require('global-agent').bootstrap();
     } else {
       // `global-tunnel-ng` works with Node.js v10 and below.
       require('global-tunnel-ng').initialize();
--- a/src/proxy.ts
+++ b/src/proxy.ts
@@ -1,25 +1,7 @@
 import * as debug from 'debug';
+import { getEnv } from './utils';
 
 const d = debug('@electron/get:proxy');
-
-/**
- * get process.env variable
- * ignore case and compatible `GLOBAL_AGENT_` namespace
- */
-function getProxyEnv(name: string): string | undefined {
-  const namespace = 'GLOBAL_AGENT_';
-  const namespaceProxyName = `${namespace}${name}`;
-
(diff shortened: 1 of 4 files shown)


src/proxy.ts
```diff
@@ -12,7 +12,9 @@ export function initializeProxy(): void {
 
     if (MAJOR_NODEJS_VERSION >= 10) {
       // `global-agent` works with Node.js v10 and above.
-      require('global-agent').bootstrap();
+      require('global-agent').bootstrap({
+        environmentVariableNamespace: '',
```
erickzhao, line 16: I don't have much context, but what's the downside of using the default `GLOBAL_AGENT_` namespace rather than the blank one here? Is there an upstream dependency that depends on having `HTTP_PROXY` as the environment variable specifically?

BlackHole1: If you using the default `GLOBAL_AGENT` namespace, the following problems will occur:

1. This is not described in https://github.com/electron/get#proxies. Leading many to think it is `HTTP(S)_PROXY` rather than `GLOBAL_AGENT_HTTP(S)_PROXY`.  This is the main reason, as most don't actually bother to read the documentation carefully. This is especially true when this is a dependency of some software. For example `electron-fiddle`. This is why, in electron issues, you will find some Chinese who cannot provide a demo because they cannot configure the proxy properly, and some "undownloadable" issues in `electron-fiddle`
2. When the nodejs version is less than 10, xxx will be used and it will use HTTP(S)_PROXY as a proxy. And we should keep our behaviour uniform

erickzhao: @BlackHole1 Since this would break existing functionality for folks who use the `GLOBAL_AGENT_` prefix, I think this would count as a major version breaking change.

If we are doing breaking changes... I wonder if we should just drop Node 10 support going forward since it's already at EOL and document the `global-agent` prefix in the README.

That would also remove our dependency to `global-tunnel-ng` :)

BlackHole1: Hey @erickzhao . Now no breaking changes. request re-reivew

src/proxy.ts
```diff
@@ -2,6 +2,25 @@ import * as debug from 'debug';
 
 const d = debug('@electron/get:proxy');
 
+/**
+ * get process.env variable
+ * ignore case and compatible `GLOBAL_AGENT_` namespace
```
malept, line 7:
```suggestion
 * Find the value of a proxy related environment variable which may or may not have the
 * prefix `GLOBAL_AGENT_`, in a case-insensitive manner.
```

src/proxy.ts
```diff
@@ -2,6 +2,25 @@ import * as debug from 'debug';
...
+
+  for (const envName in process.env) {
+    if (namespaceProxyName.toLowerCase() === envName.toLowerCase()) {
+      return process.env[envName];
+    }
+
+    if (name.toLowerCase() === envName.toLowerCase()) {
+      return process.env[envName];
+    }
+  }
```
malept, line 21: Seems like this can be simplified:
```suggestion
  const possibleKeys = [namespaceProxyName.toLowerCase(), name.toLowerCase()];
  const matchingKey = Object.keys(process.env).find((key) => possibleKeys.includes(key.toLowerCase()));

  if (matchingKey) {
    return process.env[matchingKey];
  }
```

src/proxy.ts
```diff
@@ -2,6 +2,25 @@ import * as debug from 'debug';
 
 const d = debug('@electron/get:proxy');
 
+/**
+ * get process.env variable
+ * ignore case and compatible `GLOBAL_AGENT_` namespace
+ */
+function getProxyEnv(name: string): string | undefined {
```
malept, line 9: It's probably worth adding tests for this function to avoid regressions.

malept (changes requested):

src/proxy.ts
BlackHole1: This approach would indeed be simplified, but it implicitly has three loops: `keys`, `find`, `includes`.
At the same time, I found a problem with this method: it does not respect the policy that GLOBAL_AGENT_ has a higher priority.

Let me change that

BlackHole1:
Hey @malept 👋. I have optimised the code and added unit tests.
The performance of the code is now very much faster than before and the code is also much more simplified.
request re-review

BlackHole1:
ping.

test/utils.spec.ts
```diff
@@ -155,4 +156,41 @@ describe('utils', () => {
...
+
+    it('should return has prefix environment variable if has prefix', () => {
+      const env = getEnv(prefix);
+
+      expect(env(envName)).toEqual(hasPrefixValue);
+      expect(env(envName.toLowerCase())).toEqual(hasPrefixValue);
+      expect(env(envName.toUpperCase())).toEqual(hasPrefixValue);
+    });
+
+    it('should return non-prefix environment variable if no has prefix', () => {
```
malept, line 183:
```suggestion
    it('should return non-prefixed environment variable if no prefixed variable found', () => {
```

src/proxy.ts
```diff
@@ -11,6 +12,12 @@ export function initializeProxy(): void {
     const MAJOR_NODEJS_VERSION = parseInt(process.version.slice(1).split('.')[0], 10);
 
     if (MAJOR_NODEJS_VERSION >= 10) {
+      // See: https://github.com/electron/get/pull/214
```
malept, line 15: This should link to the specific discussion.

```suggestion
      // See: https://github.com/electron/get/pull/214#discussion_r798845713
```

malept (commented):
Just a few minor things and this can be merged.

BlackHole1:
@malept done

malept (approved):

BlackHole1:
There seems to be a problem with the CI. It is in a Pendding state.

BlackHole1:
ping @erickzhao

erickzhao (approved):

erickzhao:
Not sure _why_ the tests aren't being triggered. They seemed to have worked for [e5893fc](https://github.com/electron/get/pull/214/commits/e5893fc06b1bf8be4ff14f01c874ff368f96c0ce). Maybe push an empty commit to re-kick?

BlackHole1:
> Not sure _why_ the tests aren't being triggered. They seemed to have worked for [e5893fc](https://github.com/electron/get/pull/214/commits/e5893fc06b1bf8be4ff14f01c874ff368f96c0ce). Maybe push an empty commit to re-kick?

I tried but to no avail

malept:
I think you have to re-log into CircleCI?

BlackHole1:
> I think you have to re-log into CircleCI?

I don't have access to CircleCI. So logging in again should be useless too

malept:
You should have access since you're a member of the Electron org.

BlackHole1:
> You should have access since you're a member of the Electron org.

wow. you are right. It works fine now

BlackHole1:
@malept @erickzhao Can we merge now? I'm ready to submit my next PR

Issue #215, opened 2022-03-14
GLOBAL_AGENT.HTTP_PROXY configuration value: URL protocol must be http:


zzz08900:
### Preflight Checklist

* [x] I have read the [contribution documentation](https://github.com/electron/electron-packager/blob/main/CONTRIBUTING.md) for this project.
* [x] I agree to follow the [code of conduct](https://github.com/electron/electron/blob/main/CODE_OF_CONDUCT.md) that this project follows, as appropriate.
* [x] I have searched the issue tracker for a bug that matches the one I want to file, without success.

### Issue Details

* **Electron Packager Version:**
  *  14.2.1, 15.4.0
* **Electron Version:**
  *  17.1.2
* **Operating System:**
  *  Ubuntu 20.04 x64
* **Last Known Working Electron Packager version:**
  *  13.1.1

### Expected Behavior

Building Win32 packages from my Ubuntu machine (with Wine installed) using `electron-packager . --platform=win32 --out=./builds`
Should get me Win32 artifacts in the ./builds folder.

### Actual Behavior

Building Win32 packages from my Ubuntu machine (with Wine installed) using `electron-packager . --platform=win32 --out=./builds`
results in error message: Unsupported `GLOBAL_AGENT.HTTP_PROXY` configuration value: URL protocol must be "http:".
The error originated from global-agent which is an optional dependency of @electron/get. 

### To Reproduce

1. find a Linux machine, create an empty nodeJS project, setup a barebone package.json
2. put the following in package.json 
  "devDependencies": {
    "electron": "^17.1.2",
    "electron-devtools-installer": "^3.2.0",
    "electron-packager": "^15.4.0"
  }
3. try `npx electron-packager . --platform=win32 --out=./builds`, you shoule see the error message

### Additional Information

Debug info attached below:
  electron-packager Electron Packager 15.4.0
  electron-packager Node v14.17.5
  electron-packager Host Operating system: linux 5.13.0-30-generic (x64) +0ms
  electron-packager Packager Options: {"_":["."],"platform":"win32","out":"./builds","deref-symlinks":true,"derefSymlinks":true,"download":{"rejectUnauthorized":true,"reject-unauthorized":true},"junk":true,"prune":true,"dir":".","protocols":[]} +2ms
  electron-packager Target Platforms: win32 +1ms
  electron-packager Target Architectures: x64 +0ms
  electron-packager Requiring author in package.json, as CompanyName was not specified for win32metadata +0ms
[...]

PR #216, opened 2022-03-14
fix: parse proxy error

BlackHole1:
Close: https://github.com/electron/get/issues/215

Sorry. This is a very stupid bug... 🤦‍♂️🤦‍♂️🤦‍♂️

Because:

```js
process.env.A = undefined

process.env.A === 'undefined' // => true

typeof process.env.A // => 'string'
```

cc: @malept

### DIFF
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -114,7 +114,7 @@ export function getEnv(prefix = ''): (name: string) => string | undefined {
 }
 
 export function setEnv(key: string, value: string | undefined): void {
-  if (value) {
+  if (value !== void 0) {
     process.env[key] = value;
   }
 }
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -104,11 +104,17 @@ export function getEnv(prefix = ''): (name: string) => string | undefined {
     envsLowerCase[envKey.toLowerCase()] = process.env[envKey];
   }
 
-  return (name: string) => {
+  return (name: string): string | undefined => {
     return (
       envsLowerCase[`${prefix}${name}`.toLowerCase()] ||
       envsLowerCase[name.toLowerCase()] ||
       undefined
     );
   };
+}
+
+export function setEnv(key: string, value: string | undefined): void {
+  if (value) {
+    process.env[key] = value;
+  }
 }
--- a/src/utils.ts
+++ b/src/utils.ts
@@ -114,6 +114,8 @@ export function getEnv(prefix = ''): (name: string) => string | undefined {
 }
 
 export function setEnv(key: string, value: string | undefined): void {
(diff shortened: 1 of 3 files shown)


src/utils.ts
```diff
@@ -104,11 +104,17 @@ export function getEnv(prefix = ''): (name: string) => string | undefined {
...
+  return (name: string): string | undefined => {
     return (
       envsLowerCase[`${prefix}${name}`.toLowerCase()] ||
       envsLowerCase[name.toLowerCase()] ||
       undefined
     );
   };
 }
+
+export function setEnv(key: string, value: string | undefined): void {
```
malept, line 116: Nit: I think this can be
```suggestion
export function setEnv(key: string, value?: string): void {
```

src/utils.ts
```diff
@@ -104,11 +104,17 @@ export function getEnv(prefix = ''): (name: string) => string | undefined {
...
+  if (value) {
```
malept, line 117: I'm going to assume that we're going to ignore [all falsey values in JavaScript](https://stackoverflow.com/a/19839953)?

test/utils.spec.ts
```diff
@@ -193,3 +194,17 @@ describe('utils', () => {
     });
   });
 });
+
+describe('setEnv()', () => {
+  it("don't set env", () => {
```
malept, line 199: Nit:
```suggestion
  it("doesn't set the environment variable if the value is undefined", () => {
```

test/utils.spec.ts
```diff
@@ -193,3 +194,17 @@ describe('utils', () => {
...
+    const [key, value] = ['Set_AAA_electron', undefined];
+    setEnv(key, value);
+    expect(process.env[key]).toEqual(undefined);
+  });
+
+  it('success set env', () => {
```
malept, line 205:
```suggestion
  it('successfully sets the environment variable when the value is defined', () => {
```

erickzhao (approved):

src/utils.ts
```diff
@@ -104,11 +104,17 @@ export function getEnv(prefix = ''): (name: string) => string | undefined {
...
     return (
       envsLowerCase[`${prefix}${name}`.toLowerCase()] ||
       envsLowerCase[name.toLowerCase()] ||
       undefined
     );
   };
 }
+
+export function setEnv(key: string, value: string | undefined): void {
+  if (value !== void 0) {
```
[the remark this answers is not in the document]
malept:
```suggestion
  // The `void` operator always returns `undefined`.
  // See: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/void
  if (value !== void 0) {
```

Issue #224, opened 2022-06-26
Dependabot alerts: Got allows a redirect to a UNIX socket


el3um4s:
This warning appears in my projects:

**Dependabot alerts: _Got_ allows a redirect to a UNIX socket**

_The latest possible version that can be installed is 9.6.0 because of the following conflicting dependencies:_

```
electron@19.0.6 requires got@^9.6.0 via @electron/get@1.14.1
electron-builder@23.1.0 requires got@^9.6.0 via a transitive dependency on package-json@6.5.0
nodemon@2.0.18 requires got@^9.6.0 via a transitive dependency on package-json@6.5.0
```

Is there anything that can be done?

shellatomic:
Could anyone help me to fix this issue:
Open Redirect [Medium Severity][https://snyk.io/vuln/SNYK-JS-GOT-2932019] in got@9.6.0
introduced by @electron/get@1.14.1 > got@9.6.0 and 1 other path(s) This issue was fixed in versions: 11.8.5, 12.1.

malept:
In my opinion this is not a high priority for `@electron/get`. In order to trigger this, it needs to either hijack the user's DNS or their environment variables. If either of those conditions happen, you have larger problems and the attacker might as well do other, less convoluted attacks.

That being said, it will eventually get fixed. However, due to the major version bump of `got` (which increases the minimum Node.js version), this will *not* affect any currently released version of the `electron` npm module, as changing the minimum Node.js version (even transitively) is considered a major breaking change.

kayahr:
It would be nice to get this fixed by applying the already existing PR. I use Electron in a lot of my projects and it is a bit annoying to get audit warnings from NPM and from Github's dependabot all the time because of this got issue.

In my opinion it is not a good practice to rely on old dependencies which seems to be no longer supported upstream. If you have strong reasons to stay at version 9 then maybe ask the author of got to publish a new patched version 9 to get this issue out of the way?

kayahr:
Looks like this problem is already fixed since @electron/get v2 which is used in electron v22. Problem is, electron 22 is not yet released (but installable with `npm i electron@beta`). So we just have to wait. There is nothing else to do.

And I guess this ticket can be closed.

PR #225, opened 2022-06-29
chore!: upgrade got to 11.8.5

clavin:
Bumps the dependency on `got` to version 11.8.5, fixing a dependabot alert. `got` also now includes its own type definitions, so the dependency on `@types/got` could be removed as well.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -21,13 +21,13 @@ version: 2.1
 orbs:
   win: circleci/windows@1.0.0
 jobs:
-  test-linux-8:
-    docker:
-      - image: circleci/node:8
-    <<: *steps-test
   test-linux-10:
     docker:
-      - image: circleci/node:10
+      - image: circleci/node:10.19
+    <<: *steps-test
+  test-linux-12:
+    docker:
+      - image: circleci/node:12
     <<: *steps-test
   test-mac:
     macos:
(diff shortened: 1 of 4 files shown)


malept (changes requested):
I think this is a breaking change, as it raises the minimum Node.js version?

clavin:
Ah, you're correct. The requirement for got v11 is `"node": ">=10.19.0"`, whereas this package is still `"node": ">=8.6"`. Nice catch!

malept:
A couple of things:

1. The PR title should reflect that this is a breaking change: `chore!: upgrade got to 11.8.5`
2. I think you need to re-log-into CircleCI so that the tests will run. If I had to guess it's because of the multiple-orgs-that-require-SSO problem.

clavin:
Looks like a dependency from this got version (`compress-brotli `) actually requires the min version to be `>=12`, even though `got` is itself only marked as `>=10.19`. Also have to bump the `@types/node` dependency here.

malept:
See https://github.com/electron/get/issues/224#issuecomment-1177920828, merging this change needs to be coordinated properly.

BlackHole1:
Based on https://github.com/electron/get/pull/226#issuecomment-1208824651 discussions. I think we should just upgrade instead of waiting for `got` to release a new patch version

malept:
We're not waiting for `got` to release a new patch version.

BlackHole1:
What are the reasons that are stopping this PR from being merged now? I want to push for this PR to be merged.

The most important reason, as I understand it so far, is the impact of the minimum node version

BlackHole1:
If this impact is so important to us that we cannot merge this PR in a short time, is there another way, for example if we fork `got`

malept (approved):

PR #226, opened 2022-08-08
fix!: download electron checksum failure

BlackHole1:
Bug context see: https://github.com/electron/fiddle/issues/1156

> This bug is only triggered in the `electron-renderer` and when `Webpack` is packaged.

## Break Changes

1. Node.js version must >= `12.0.0`

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -21,13 +21,13 @@ version: 2.1
 orbs:
   win: circleci/windows@1.0.0
 jobs:
-  test-linux-8:
+  test-linux-12:
     docker:
-      - image: circleci/node:8
+      - image: circleci/node:12
     <<: *steps-test
-  test-linux-10:
+  test-linux-14:
     docker:
-      - image: circleci/node:10
+      - image: circleci/node:14
     <<: *steps-test
   test-mac:
     macos:
@@ -53,11 +53,13 @@ workflows:
     # Run the test jobs first, then the release only when all the test jobs are successful
     jobs:
       - test-linux-12
+      - test-linux-14
       - test-mac
       - test-windows
       - release:
           requires:
             - test-linux-12
+            - test-linux-14
             - test-mac
             - test-windows
           filters:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -52,14 +52,12 @@ workflows:
   test_and_release:
     # Run the test jobs first, then the release only when all the test jobs are successful
     jobs:
-      - test-linux-8
-      - test-linux-10
+      - test-linux-12
       - test-mac
       - test-windows
       - release:
           requires:
-            - test-linux-8
-            - test-linux-10
+            - test-linux-12
             - test-mac
             - test-windows
           filters:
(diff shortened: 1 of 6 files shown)


BlackHole1:
CI failed reason: `Cannot find a definition for job named test-linux-12`. cc @malept

.circleci/config.yml
```diff
@@ -52,14 +52,12 @@ workflows:
   test_and_release:
     # Run the test jobs first, then the release only when all the test jobs are successful
     jobs:
-      - test-linux-8
-      - test-linux-10
+      - test-linux-12
```
erickzhao, line 55: For this to be defined, you should define a `test-linux-12` job as well.

https://github.com/electron/get/blob/c49bf7f30801917c7c59a1be56066e5fc7f69dd6/.circleci/config.yml#L28-L31

erickzhao (changes requested):

malept (commented):
Can you explain why this change requires `got` to be updated? This seems like a pretty large change for a single bug.

BlackHole1:
It may take time to explain this matter from beginning to end, so please wait. In the meantime, you can start by checking out: https://github.com/electron/fiddle/issues/1156

BlackHole1:
## Tips
1. It is difficult to write the corresponding unit tests in `electron/get`. Because the problem only occurs in the `electron-renderer` process and uses the `Webpack` build. This can be reproduced at https://github.com/electron/fiddle/releases/tag/v0.29.2 (Select China Mirrors)
2. This bug is fixed at https://github.com/sindresorhus/got/pull/1105
3. There is a discussion of this bug in `got`. https://github.com/sindresorhus/got/issues/945

### Why update `got`

Yesterday, I found that my electron-fiddle was not downloading electron, and after some troubleshooting, I figured out that it was not parsing `gzip` when downloading with `got.stream`. So when `@electron/get` uses `got.stream` to download `SHASUMS256.txt`, the response returns gzip, which causes `@electron/get` to fail the checksum.

https://github.com/electron/get/blob/5c81f9a388577a9d446b2f7ae1a6e2dd2d7177d6/src/index.ts#L131-L141

https://github.com/electron/get/blob/5c81f9a388577a9d446b2f7ae1a6e2dd2d7177d6/src/index.ts#L151-L153

https://github.com/electron/get/blob/5c81f9a388577a9d446b2f7ae1a6e2dd2d7177d6/src/GotDownloader.ts#L60

### This seems like a pretty large change for a single bug.

We talked about "no longer supporting node.js <  10" once before. See: https://github.com/electron/get/pull/214#discussion_r804004094

Because we have other problems caused by the low version of got https://github.com/electron/get/issues/224

**I think this is the time to do it.**

malept:
It would be helpful if you could split out the changes into separate logical commits so that it's easier to review.

.circleci/config.yml
BlackHole1: I'm not very familiar with `circleci`, thanks for the correction

BlackHole1:
> It would be helpful if you could split out the changes into separate logical commits so that it's easier to review.

Sorry, I didn't quite understand it. This PR is already a minimal change and cannot be split any further

malept:
> > It would be helpful if you could split out the changes into separate logical commits so that it's easier to review.
> 
> Sorry, I didn't quite understand it. This PR is already a minimal change and cannot be split any further

You can start with #225 and build upon it.

BlackHole1:
OK, I will close this PR as there are duplicate PRs

PR #227, opened 2022-08-11
fix: remove global-tunnel-ng library

BlackHole1:
As of v2.0.0, the minimum nodejs version we support is 12. so `global-tunnel-ng` no longer has any effect.

cc: @malept


malept:
This should be a fix since it's removing an unnecessary dependency

BlackHole1:
> This should be a fix since it's removing an unnecessary dependency

Done

erickzhao (approved):

BlackHole1:
Looks like we can merge?

PR #229, opened 2022-08-11
chore: ignore .idea directory in git

BlackHole1:


malept (commented):
I'd argue that either you put this in your global `.gitignore`, or we ignore all editors' specific directories/files. I'm generally more in favor of the former.

BlackHole1:
> I'd argue that either you put this in your global `.gitignore`, or we ignore all editors' specific directories/files. I'm generally more in favor of the former.

It may not be a good idea to put it globally, as the `.idea` has a role in some projects (e.g. debug). See: https://github.com/netless-io/flat/blob/main/docs/debugging/electron/README.md#webstorm

Although `.idea` and `.vscode` are the same in nature, there are many more files in `.idea` than in `.vscode`, which is the main reason why many projects don't put `.idea` in git.

I can submit another PR to migrate the vscode configuration from the project to the `.editorconfig` file.

VerteDinde (approved):

PR #230, opened 2022-08-29
feat: add environment variable GITHUB_TOKEN support.

mrchaofan:
I met a trouble when I executed yarn install. 
image
It seems like github blocked but if I add github token in request header it would get fixed.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -81,6 +81,16 @@ export async function downloadArtifact(
     }
   }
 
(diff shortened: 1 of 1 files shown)


MarshallOfSound (changes requested):
We don't hit any APIs afaik, not sure why you'd be getting 429'ed

We shouldn't do this without more thought, currently this code would send your GitHub token to any configured mirror as well :/

PR #234, opened 2022-12-01
fix: update dependencies to clean up yarn audit

MarshallOfSound:
BREAKING CHANGE: The minimum supported node version is now node 14

This PR also as a side effect ensures tests can pass / run on M1 hardware.


BlackHole1:
CI failed. Reason: `error @continuous-auth/semantic-release-npm@3.0.0: The engine "node" is incompatible with this module. Expected version ">=16 || ^14.17". Got "12.13.1"`.

Restoration Solutions:
1. Use `--ignore-engines` In CI. Since only `devDependencies` are upgraded, it does not affect the end-users.
2. Fallback `@continuous-auth/semantic-release-npm` version
3. Modifdy https://github.com/electron/get/blob/main/package.json#L26 to `>=16`

BlackHole1:
CI failed again.
image

Optional Chaining was not implemented until after nodejs version 14

package.json
```diff
@@ -97,4 +96,4 @@
     "eslint/inquirer": "< 7.3.0",
     "**/@typescript-eslint/typescript-estree/semver": "^6.3.0"
   }
-}
+}
```
BlackHole1, line 99:
```suggestion
}

```

package.json
```diff
@@ -19,11 +19,11 @@
     "test:nonetwork": "npm run lint && npm run jest -- --testPathIgnorePatterns network.spec"
   },
   "files": [
-    "dist/*",
+    "dist/",
```
BlackHole1, line 22: Currently, many libraries publish their src to npm, which is very useful for developers during the debugging phase. Therefore, my suggestion is that we also publish our src.

```suggestion
    "dist/",
    "src/",
```

BlackHole1 (dismissed):

erickzhao (approved):

package.json
MarshallOfSound: I don't think we should publish src, it can confuse things like typescript and the whole point of an npm package is you publish the smallest amount of files required for your package to work.  None of our other packages publish `src`, this isn't going to be the first one

dsanders11: Forge packages currently publish `src`, actually (see https://github.com/electron/forge/pull/3284). The rationale at the time is we're shipping source maps which don't do any good if there's not also the `src` directory they reference.

I don't have a strong opinion on this either way, but I think if we don't publish `src`, we shouldn't publish source maps in `dist` either.

PR #280, opened 2024-04-07
Download Progress Bar

botPuneet:
Created PR to resolve issue #[issue](https://github.com/electron/fiddle-core/issues/59)


dsanders11:
Thanks for the PR, but I don't think this is the right approach to solving electron/fiddle-core#59. @marfgold1 has shown the bug is due to redirects and invalid values for `total`. This PR would lead to `NaN` in those cases.

botPuneet:
Sorry @dsanders11 sir i have misunderstood the problem. thank u for the feedback. just one question is this issue still open?

> Thanks for the PR, but I don't think this is the right approach to solving [electron/fiddle-core#59](https://github.com/electron/fiddle-core/issues/59). @marfgold1 has shown the bug is due to redirects and invalid values for `total`. This PR would lead to `NaN` in those cases.

PR #288, opened 2024-06-11
docs: improve types

erickzhao:
Cleanup of our API documentation for `@electron/get`.

Notable changes:
* Extracted `ElectronGenericArtifactDetails` out of `ElectronPlatformArtifactDetailsWithDefaults`.
* Typedoc upgrade requires Node.js 16 (can revert if we don't want this)

Other work is mostly categorizing, expanding on existing comments, and hiding internal types.


.vscode/settings.json
```diff
@@ -1,5 +0,0 @@
-{
-    "editor.formatOnSave": true,
```
erickzhao, line 2: Feel like this should not have been committed to code. The tab width and quote styles are already covered via Prettier configuration.

package.json
```diff
@@ -52,7 +52,7 @@
     "lint-staged": "^13.0.4",
     "prettier": "^1.17.1",
     "ts-jest": "^29.0.0",
-    "typedoc": "^0.23.21",
+    "typedoc": "~0.25.13",
```
erickzhao, line 55: This is an optional update for this PR and actually breaks CI because we support Node.js 14 in this package.

src/index.ts
```diff
@@ -99,16 +100,22 @@ async function validateArtifact(
...
+ * platform/arch-specific (e.g. `ffmpeg-v31.0.0-darwin-arm64.zip`) and others that
+ * are generic (e.g. `SHASUMS256.txt`).
+ *
+ *
  * @param artifactDetails - The information required to download the artifact
+ * @category Download Artifact
  */
 export async function downloadArtifact(
-  _artifactDetails: ElectronPlatformArtifactDetailsWithDefaults,
+  artifactDetails: ElectronPlatformArtifactDetailsWithDefaults | ElectronGenericArtifactDetails,
```
erickzhao, line 112: Doesn't make sense to have the function parameter start with `_` so I changed the variable naming around.

src/index.ts
```diff
@@ -148,40 +155,43 @@ export async function downloadArtifact(
...
+    await validateArtifact(details, tempDownloadPath, downloadArtifact);
 
     return await cache.putFileInCache(url, tempDownloadPath, fileName);
   });
 }
 
 /**
- * Downloads a specific version of Electron and returns an absolute path to a
+ * Downloads the Electron binary for a specific version and returns an absolute path to a
  * ZIP file.
```
erickzhao, line 189: Technically, we have some non-ZIP artifacts too as well? such as SHASUM256.txt

package.json
erickzhao: downgraded to 0.24 in [fe2dfa6](https://github.com/electron/get/pull/288/commits/fe2dfa652a272c84e55cb1ed19c02240641689b5)

erickzhao: this is still incompatible because Node doesn't have v14 binaries for arm64, so we hacked CI to use Node 14.16... and `typedoc 0.24` requires Node.js >= 14.17.

I think there are actually a lot of features in TypeDoc 0.24 that we need. :(

erickzhao: As discussed over video call earlier this week, `--ignore-engines` applied for Node.js 14 CI.

georgexu99 (approved):
looks good, I'm excited to see more types in our docs :D

PR #312, opened 2025-01-11
feat!: bump engines to Node.js >=22.12.0

erickzhao:
BREAKING CHANGE: bumps minimum Node.js version to >=22.12.0

## ESM

Node.js 22.12 gives us the ability to use CommonJS `require` calls on ESM dependency graphs. This compatibility gives us the ability to migrate our libraries to ESM and maintain support for projects that are using CommonJS.

## Removal of deprecated APIs

`@electron/get` already had a few deprecated APIs and this PR removes them since we're already making a breaking change anyways:

* `nightly_mirror` option was removed (was deprecated in favour of `nightlyMirror`)
* `force` option was removed (was deprecated in favour of `cacheMode=WriteOnly`).

## Test migration from Jest to Vitest

Jest currently has issues with `require(esm)` while it works out of the box with Vitest. We were already looking to migrate to Vitest across other Ecosystem repos, so this PR just switches over.

* Need to import functions from `vitest` (no globals)
* `jest.mock` calls turned into `vi` calls
* Regenerated snapshots

## tsconfig.json changes

We're now basing our TypeScript config off of `@tsconfig/node22`, which is a bit different from what we had in the past.

Due to going ESM-only, we only have a singular `tsconfig.json` now.


src/Cache.ts
```diff
@@ -39,12 +40,16 @@ export class Cache {
   public async putFileInCache(url: string, currentPath: string, fileName: string): Promise<string> {
     const cachePath = this.getCachePath(url, fileName);
     d(`Moving ${currentPath} to ${cachePath}`);
-    if (await fs.pathExists(cachePath)) {
-      d('* Replacing existing file');
-      await fs.remove(cachePath);
+
+    if (!fs.existsSync(path.dirname(cachePath))) {
```
erickzhao, line 44: Note: this new clause is needed because `fs.rename` will fail if the target directory doesn't exist, while `fs.move` creates the directory.

This also doesn't work across disks, but we could catch the resulting error and fall back to an `fs.copy` call if necessary.

package.json
```diff
@@ -14,89 +14,55 @@
...
+    "node": ">=22.0.0"
   },
   "dependencies": {
     "debug": "^4.1.1",
     "env-paths": "^2.2.0",
-    "fs-extra": "^8.1.0",
     "got": "^11.8.5",
     "progress": "^2.0.3",
     "semver": "^6.2.0",
     "sumchecker": "^3.0.1"
```
dsanders11, line 34: Do we want to bump major versions here while we're at it?

erickzhao: Makes sense to do so to the best of our abilities. I'll give it a shot!

test/artifact-utils.spec.ts
```diff
@@ -109,8 +111,7 @@ describe('artifact-utils', () => {
           artifactName: 'electron',
           mirrorOptions: {
             mirror: 'https://mirror.example.com/',
-            // eslint-disable-next-line @typescript-eslint/camelcase
```
felixrieseberg, line 112: Thank god

tsconfig.json
```diff
@@ -1,22 +1,9 @@
...
     "sourceMap": true,
-    "strict": true,
-    "outDir": "dist/cjs",
-    "types": [
-      "node",
-      "jest"
-    ],
-    "allowSyntheticDefaultImports": true,
-    "moduleResolution": "node",
-    "declaration": true
```
erikian, line 17: question: `declaration: true` is not included in the base `@tsconfig/node22/tsconfig.json` config, is the removal intentional here?

dsanders11: @erikian, good catch! I definitely think we want to keep `declaration: true` here.

dsanders11:
> and uses `util.promisify` because `graceful-fs` unfortunately doesn't patch `fs.promises`

Sad, but good catch!

VerteDinde (approved):
Excited to get this landed!

PR #317, opened 2025-02-19
docs: update API documentation link in README

erickzhao:
Now that https://packages.electronjs.org/get/v4.0.0/index.html is live :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@
 
 ## Usage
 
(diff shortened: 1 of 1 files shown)


dsanders11 (approved):
LGTM, contingent on this being temporary until we get `latest` redirects working so we don't have to hardcode the version in the URL.

Issue #332, opened 2025-05-05
"Error: EXDEV: cross-device link not permitted" on version 4.0.0


weedz:
### System

Operating system: Linux 6.14.4
`/tmp` mounted as a "tmpfs". `/home` mounted as a regular ext4 partition.

### Description

Seems like when `fs-extra` was replaced by `graceful-fs`, the `fs.move()` here <https://github.com/electron/get/blame/ee6c19787305e22b5cb5ca90b3c9764ca2a4867b/src/Cache.ts#L47> was replaced with `fs.promises.rename()` (<https://github.com/electron/get/compare/v3.1.0...v4.0.0#diff-b9f2cbd9f2764d566a3e7e68b3a5893cd2ec3d05f3b81c26edeeb203fc1e760eR52>) which does not handle cross-device links.

The following code:
```js
import { downloadArtifact } from "@electron/get";

const version = "36.0.0";

downloadArtifact({
  version,
  artifactName: 'electron',
  platform: "linux",
  arch: "x64",
}).then(filename => {
  console.log("File:", filename);
}).catch(err => {
  console.error(err.stack);
  process.exit(1);
});
```
exits with the following error:
```
Error: EXDEV: cross-device link not permitted, rename '/tmp/electron-download-miB9bq/electron-v36.0.0-linux-x64.zip' -> '/home/user/.cache/electron/ad19b0ca2284b4694c510be19fa6fb0edc857d005acb0aa9a584e244f99a3e4b/electron-v36.0.0-linux-x64.zip'
    at async Object.rename (node:internal/fs/promises:778:10)
    at async Cache.putFileInCache (file:///home/user/workspace/electron-get/node_modules/.pnpm/@electron+get@4.0.0/node_modules/@electron/get/dist/Cache.js:43:9)
    at async file:///home/user/workspace/electron-get/node_modules/.pnpm/@electron+get@4.0.0/node_modules/@electron/get/dist/index.js:161:20
    at async useAndRemoveDirectory (file:///home/user/workspace/electron-get/node_modules/.pnpm/@electron+get@4.0.0/node_modules/@electron/get/dist/utils.js:9:18)
    at async downloadArtifact (file:///home/user/workspace/electron-get/node_modules/.pnpm/@electron+get@4.0.0/node_modules/@electron/get/dist/index.js:149:12)
```

On version 3.1.0 the program runs successfully and print:
```
[...]

PR #335, opened 2025-05-19
fix: cache works properly when /tmp is on a separate file system

cacieprins:
Fix for https://github.com/electron/get/issues/332.

When `/tmp` is on a different file system than the cache, `fs.promises.rename` will throw an error in linux.

This change uses `copyFile` and `rm` rather than `rename`, and introduces a unit test to verify that `putFileInCache` still succeeds if `rename` throws an error.

### DIFF
--- a/src/Cache.ts
+++ b/src/Cache.ts
@@ -49,7 +49,9 @@ export class Cache {
       d('* Replacing existing file');
       await fs.promises.rm(cachePath, { recursive: true, force: true });
     }
-    await fs.promises.rename(currentPath, cachePath);
+
+    await fs.promises.copyFile(currentPath, cachePath);
+    await fs.promises.rm(currentPath);
 
     return cachePath;
   }
--- a/test/Cache.spec.ts
+++ b/test/Cache.spec.ts
@@ -3,7 +3,7 @@ import os from 'node:os';
 import path from 'node:path';
 import util from 'node:util';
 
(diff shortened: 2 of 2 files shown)


src/Cache.ts
```diff
@@ -49,7 +49,9 @@ export class Cache {
       d('* Replacing existing file');
       await fs.promises.rm(cachePath, { recursive: true, force: true });
     }
-    await fs.promises.rename(currentPath, cachePath);
+
+    await fs.promises.copyFile(currentPath, cachePath);
+    await fs.promises.rm(currentPath);
```
BlackHole1, line 54: The performance of rename is better than cp + rm. We prioritize using rename, and only when rename fails with an error of EXDEV do we fallback to cp + rm.

```suggestion
    try {
      await fs.promises.rename(currentPath, cachePath);
    } catch (err) {
      if ((err as NodeJS.ErrnoException).code !== "EXDEV") {
        throw err;
      }
      
      await fs.promises.cp(currentPath, cachePath, { recursive: true, force: true });
      await fs.promises.rm(currentPath, { recursive: true, force: true });
    }
```

BlackHole1 (changes requested):

dsanders11:
This has been resolved in #336, but thank you for the PR @cacieprins! We appreciate it. 👍

PR #337, opened 2025-07-17
fix: avoid permission errors on windows during double-installs

MarshallOfSound:
Double installs sometimes fight each other for the cache entry, let's back off a bit and then check if the thing in the cache matches the thing we're putting in and if so just accept it.

### DIFF
--- a/package.json
+++ b/package.json
@@ -80,5 +80,6 @@
   ],
   "optionalDependencies": {
     "global-agent": "^3.0.0"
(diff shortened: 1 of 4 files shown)


felixrieseberg (approved):

VerteDinde (approved):

dsanders11 (changes requested):
Looks like there's a test failure.

Issue #243, opened 2023-04-07
The "Got" library failed to download electron behind local proxy


lucascloarec:
Hi,
I use winfoom (https://github.com/ecovaci/winfoom) as a local proxy to connect my apps to our corporate proxy and I encounter the following error:

```text
npm ERR! code 1
npm ERR! path C:\...\workspace\pki-gnokey\node_modules\electron
npm ERR! command failed
npm ERR! command C:\Program Files\Git\bin\bash.exe -c node install.js
npm ERR! RequestError: write EPROTO DC7D0000:error:0A00010B:SSL routines:ssl3_get_record:wrong version number:c:\ws\deps\openssl\openssl\ssl\record\ssl3_record.c:355:
npm ERR!
npm ERR!     at ClientRequest.<anonymous> (C:\My Program Files\workspace\pki-gnokey\node_modules\got\dist\source\core\index.js:970:111)
npm ERR!     at Object.onceWrapper (node:events:628:26)
npm ERR!     at ClientRequest.emit (node:events:525:35)
npm ERR!     at ClientRequest.emit (node:domain:489:12)
npm ERR!     at origin.emit (C:\...\workspace\pki-gnokey\node_modules\@szmarczak\http-timer\dist\source\index.js:43:20)
npm ERR!     at TLSSocket.socketErrorListener (node:_http_client:502:9)
npm ERR!     at TLSSocket.emit (node:events:525:35)
npm ERR!     at TLSSocket.emit (node:domain:489:12)
npm ERR!     at emitErrorNT (node:internal/streams/destroy:151:8)
npm ERR!     at emitErrorCloseNT (node:internal/streams/destroy:116:3)
npm ERR!     at WriteWrap.onWriteComplete [as oncomplete] (node:internal/stream_base_commons:94:16)
```
I have these environment variables:
ELECTRON_GET_USE_PROXY=true
GLOBAL_AGENT_HTTP_PROXY=http://127.0.0.1:3128
http_proxy=http://127.0.0.1:3128
https_proxy=http://127.0.0.1:3128

I have the impression that the SSL/TLS negotiation does not succeed because "got" forces the use of a too old version which is not compatible with my local proxy.
Also I don't encounter the problem when I use my own implementation of the Downloader interface, with axios instead of "got" :

```typescript
import {Downloader} from "@electron/get/dist/esm/Downloader";
import * as fs from 'fs';
import axios, {AxiosRequestConfig, AxiosResponse} from "axios";
import * as Stream from "stream";


[...]

PR #368, opened 2026-02-27
refactor!: replace global-agent with proxy-agent

### LINKED ISSUE #304: Electron seems depending an unsupported package: npm warn deprecated boolean@3.2.0: Package no longer supported.
Forwarding this issue from here : [https://github.com/electron/electron/issues/44416](https://github.com/electron/electron/issues/44416)

Just encountered the same issue.

BlackHole1:
Fixed: #304 #243

The `global-agent` dependency is in a state of maintenance stagnation. Although 4 releases have been published in the three weeks ago., they are all in an unusable state (which reduces my confidence in this dependency).

Also, personally, I don't think the proxy should pollute the entire global scope - all requests are being proxied. It should be more atomic.

Fortunately, we have set the minimum node version requirement to >= 22.12.0. Based on this, we can gracefully use `proxy-agents` to use a proxy for specific requests without polluting the global scope.

Regarding proxy behavior, I referred to curl: automatically detect proxy variables and apply them. Instead of controlling whether to use a proxy through special env variables, this will reduce the cognitive load on developers (because many developers simply don't know that `npm install electron` actually calls `@electron/get` to download, let alone look at the `@electron/get` documentation).

**No Proxy ENV**
图片

**Has Proxy ENV**
CleanShot 2026-02-27 at 11 37 45

BREAKING CHANGE: `GLOBAL_AGENT_*` env vars no longer work. `ELECTRON_GET_USE_PROXY` is no longer checked. `getEnv()` and `setEnv()` removed from public API.


dsanders11:
@BlackHole1, what's the status of this after #376?

BlackHole1:
@dsanders11 I didn't notice the status of 376, I just noticed it was merged. I have now closed this PR.

